PR #29, opened 2019-11-01
[5/5] Contracts with paramaters that have names that are rust keywords cause compile errors

nlordell:
This happens for example with solidity libraries where parameters are called `self`. Also adds sample library contract that exhibits this problem. We solve this by adding a `_` to the end of the identifier.

### DIFF
--- a/derive/src/lib.rs
+++ b/derive/src/lib.rs
@@ -11,10 +11,9 @@ use ethcontract_common::truffle::Artifact;
 use inflector::Inflector;
 use proc_macro2::{Ident, Literal, Span, TokenStream};
 use quote::quote;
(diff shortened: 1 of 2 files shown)


fleupold (approved):
Cool! Maybe a unit test would be good to document this? Also it would have been awesome to shortly describe how you fix it in the PR description (appending a _ to parameters that fail to parse).

nlordell:
@fleupold I can add a unit test once I figure out how to unit test proc macros. Note that we have some sort of "test" since we generate the bindings for the contract `SimpleLibrary` which runs into this problem. Since we build with the `example` feature on the CI, we would fail to build without this fix.

Added description to PR

PR #30, opened 2019-11-04
Rename `example` feature to `samples`

### LINKED ISSUE #27: [Refactor] Rename `example` feature to something more meaningful
We currently feature gate generating bindings to sample contracts on a feature called `example`. The idea was to build documentation on `docs.rs` with this feature enabled but have it disabled by default. This would basically allow us to have `rustdoc` documentation for a sample contract as a "here is what a generated contract binding looks like".

While this works the feature name is unclear and can be confused with cargo examples which it has nothing to do with.

nlordell:
Rename the `example` feature to something else that is more descriptive and can't be confused with Cargo examples (as the feature is unrelated to Cargo examples).

also snuck in a small change to remove an unused dependency

Closes #27

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,4 +22,4 @@ script:
   - cargo build --all-features --verbose --all
   - cargo test --all-features --verbose --all
   - cargo fmt --all -- --check
-  - cargo clippy --all
+  - cargo clippy --all --all-features --all-targets -- -D warnings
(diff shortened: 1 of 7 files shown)


fleupold (approved):
<a href="https://gitme.me/image?url=https%3A%2F%2Fmedia3.giphy.com%2Fmedia%2FRl9Yqavfj2Ula%2Fgiphy.gif&token=easy" data-gitmeme-token="easy"></a>

PR #32, opened 2019-11-04
Cache `$HOME/.npm` to speed up build of example contracts

nlordell:
:point_up: 

Note that we started using `npm ci` instead of `npm install` and we cache the NPM registry at `$HOME/.npm`. Accoring to [this article](https://medium.com/hackernoon/how-to-speed-up-continuous-integration-build-with-new-npm-ci-and-package-lock-json-7647f91751a) it can really speed up your builds. Note that we can't use `cache: npm` directly since the NPM project is not in the root of our repo.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,4 +4,3 @@ Cargo.lock
 
 node_modules/
 examples/*/build/
-package-lock.json
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,12 +10,13 @@ jobs:
 
 cache:
   cargo: true
-  npm: true
+  directories:
+    - $HOME/.npm
 
 before_install:
   - rustup component add clippy rustfmt
 install:
-  - (cd examples/truffle; npm install)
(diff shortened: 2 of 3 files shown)


fleupold (approved):
@bh2smith we might want to use this as well for our contracts and services CI. 

However, I thought that even though it doesn't have to hit the network again I thought that installing all dependencies again still can take somewhat significant time for larger projects, which is why caching the node_modules directory is better.

What are your thoughts?

nlordell:
@fleupold @bh2smith I checked with Ben the other day and this is what contracts CI is currently doing as it is the default behaviour for Travis for a NodeJS project. This is what was doing an NPM install even if it was never explicitly called in the Travis file.

It might be worth investigating caching the `node_modules` as well. If I had to guess I would say the time comes from the `postinstall` scripts that might be building C/C++ code (some of the crypto packages for example) in which case caching the `node_modules` might help.

PR #34, opened 2019-11-07
[Easy] Small changes needed to get cargo publish to work

nlordell:
Just some small alterations to some of the crate metadata for publishing on crates.io.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -7,6 +7,7 @@ license = "MIT"
 repository = "https://github.com/gnosis/ethcontract-rs"
 homepage = "https://github.com/gnosis/ethcontract-rs"
 documentation = "https://docs.rs/ethcontract"
+keywords = ["web3", "ethereum", "contract", "async"]
 description = """
 Runtime library and proc macro for interacting and generating type-safe bindings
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -7,6 +7,7 @@ license = "MIT"
 repository = "https://github.com/gnosis/ethcontract-rs"
 homepage = "https://github.com/gnosis/ethcontract-rs"
 documentation = "https://docs.rs/ethcontract"
+keywords = ["web3", "ethereum", "contract", "async"]
```
bh2smith, line 10: perhaps "smart contract" is more precise? Do we want to include "solidity" as well?

bh2smith (approved):
Looks fantastic! Rock on!

Cargo.toml
nlordell: I don't want to include `solidity` since it never interfaces directly with it. Maybe `truffle`?

nlordell: I'll add it as a separate PR later when I do some cleanup (like adding badges and getting crates.io to show a better landing page)

nlordell: Created #35

PR #38, opened 2019-11-08
Async await

nlordell:
Since async/await is now stabilized, we no longer need rust version guards on the example. Note that async tests are still unstable, so we're not 100% there yet.

### Test Plan
CI


bh2smith (approved):
Safe to assume that many of the recent changes have something to do with the recent release of `1.39.0`?

PR #39, opened 2019-11-12
New Contract Method API

nlordell:
When first designing the contract API I made a wrong assumption that non-view functions could never be called and only could send transactions. This PR refactors the `CallBuilder` and `SendBuilder` API into a `ViewMethodBuilder` and `MethodBuilder` where `MethodBuilder` (which used to be the `SendBuilder`) can now also just call to be evaluated on the node without modifying state.

Note that the generated code still restricts view methods to not be able to actually send transactions.

### Test Plain
Fix existing unit tests and examples and ensure they still run.


src/contract.rs
```diff
@@ -2,21 +2,19 @@
...
 
-mod call;
 mod deploy;
-mod send;
+mod method;
 
 use crate::errors::{DeployError, LinkError};
 use crate::truffle::{Abi, Artifact, Bytecode};
-use ethabi::{Function, Result as AbiResult};
+use ethabi::{Result as AbiResult};
```
bh2smith: This is what your linter is failing!

```suggestion
use ethabi::Result as AbiResult;
```

derive/src/lib.rs
```diff
@@ -77,7 +78,10 @@ fn expand_contract(args: ContractArgs) -> Result<TokenStream> {
     //   the path will always be rooted on the cargo manifest directory.
     //   Eventually we can use the `Span::source_file` API to have a better
     //   experience.
-    let artifact_path = args.artifact_path;
+    let artifact_path = {
+        let full_path = fs::canonicalize(args.artifact_path.value())?;
+        LitStr::new(&full_path.to_string_lossy(), args.artifact_path.span())
+    };
```
fleupold, line 84: Unrelated to the purpose of this PR 🙈 

<a href="https://gitme.me/image?url=https%3A%2F%2Fmedia0.giphy.com%2Fmedia%2FB3nLKeJ9wUPbG%2Fgiphy.gif&token=unrelated" data-gitmeme-token="unrelated"></a>

src/contract/method.rs
```diff
@@ -0,0 +1,214 @@
+//! Implementation for a contract method builder and call future. This is not
```
fleupold, line 1: If you use git mv it should be able to track and only highlight the changes between the files instead of saying all lines are added.

src/contract/deploy.rs
```diff
@@ -102,6 +103,7 @@ where
 
 /// Builder for specifying options for deploying a linked contract.
 #[derive(Debug, Clone)]
+#[must_use = "deploy builers do nothing unless you `.deploy()` them"]
```
fleupold, line 106: Nice addition but again probably sh/could have been a separate PR.

fleupold (approved):

derive/src/lib.rs
nlordell: Hehe yes it is.

src/contract/method.rs
nlordell: :open_mouth:

src/contract/deploy.rs
nlordell: Yes, the issue was that I wanted to add it to the `MethodBuilder` and `ViewMethodBuilder` types as well and creating it as a separate PR would have been annoying.

nlordell: I actually added it to the MethodBuilder types and then noticed that we didn't have it on our other builder types so added it as well.

derive/src/lib.rs
nlordell: @fleupold Want me to move this to a separate PR? I agree it should probably be kept separate.

bh2smith: What does the `Lit` stand for in `LitStr`?

bh2smith (approved):
At this point in my rust career, I must say that a lot of the changes made here are beyond me. I will approve this stuff for the time being, but if you would like a proper review, I might have to spend quite a bit longer.

derive/src/lib.rs
fleupold: I think it's fine to keep it in here, but it should become a habit to create small PRs for small fixes. It makes it easier to track what changed when and allows reverting individual features. E.g. if we were to revert this commit, we'd also revert all the great `must use` annotations.

PR #64, opened 2019-12-04
Work Around Estimate Gas Issue With Geth Nodes

### LINKED ISSUE #47: Estimate Gas Does Not Work On Rinkeby with Infura
Calling `eth_estimateGas` on Infura causes a JSONRPC error with message `"too many arguments, want at most 1"`.

nlordell:
This implements a workaround to tomusdrw/rust-web3#290. While the issue has been fixed on master, it has not been released to crates.io.

Fixes #47 

### Test Plan

CI

### DIFF
--- a/examples/rinkeby.rs
+++ b/examples/rinkeby.rs
@@ -49,7 +49,6 @@ async fn run() {
     instance
         .increment()
         .from(account.clone())
-        .gas(1_000_000.into())
         .send_and_confirm(Duration::new(5, 0), 1)
         .await
         .expect("increment failed");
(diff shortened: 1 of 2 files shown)


fleupold (commented):
Would it be better to change the cargo dependency to point at the git repo instead?

nlordell:
> Would it be better to change the cargo dependency to point at the git repo instead?

Then we can't publish to crates.io :cry:

I think there are enough changes to publish 0.2.0 soon, and then we can point to latest `rust-web3` by pointing to the git repo so that we incorporate the PRs that were merged into `rust-web3` (like the estimate gas fix, the signing, the better type (de)tokenization, etc.)

My goal was to use 0.2.0 for dex-services and it would be nice if it were a published crate

fleupold (approved):
Makes sense!

PR #73, opened 2019-12-16
Method Result Not Required To Be Tokenize

nlordell:
This PR fixes generating code for solidity functions that don't return anything.

### Test Plan

dex-services

### DIFF
--- a/src/contract.rs
+++ b/src/contract.rs
@@ -98,7 +98,6 @@ impl<T: Transport> Instance<T> {
     where
         S: AsRef<str>,
         P: Tokenize,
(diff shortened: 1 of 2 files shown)


fleupold (approved):
Why was Detokenize in there in the first place if no line of code relied on detokenizing the result?

nlordell:
@fleupold It's so we can downgrade a `MethodBuilder` to a `ViewMethodBuilder` in a type-safe way. It is used in `MethodBuilder::call` and `MethodBuilder::view` and the type needs to be carried around by the `MethodBuilder` to be used those places.

PR #75, opened 2019-12-17
Added Some Re-Exports to Make Things More Convenient

nlordell:
This should allow ethcontract to be used without having to pull in any extra crates. Specifically re-exporting the `ethsign` crate is important to achieve this. Additionally some commonly used types when interacting with contracts are re-exported in the root.

### Test Plan

Compiler


fleupold (approved):

common/src/truffle.rs
```diff
@@ -10,7 +10,7 @@ use std::path::Path;
 use web3::types::Address;
 
 pub use self::bytecode::Bytecode;
-pub use ethabi::Contract as Abi;
+pub use ethabi::{self as abi, Contract as Abi};
```
bh2smith, line 13: This is weird... so we have capitalized and lowecase `abi`...

PR #78, opened 2019-12-18
[Easy] Allow Functinos With too Many Arguments and Complex Types

nlordell:
This just disables some clippy warnings on generated code for things that cannot be changed (like functions with too many arguments or return values).

### Test Plan

dex-services

### DIFF
--- a/generate/src/contract.rs
+++ b/generate/src/contract.rs
@@ -77,6 +77,7 @@ pub(crate) fn expand_contract(args: &Args) -> Result<TokenStream> {
         }
 
         #[allow(dead_code)]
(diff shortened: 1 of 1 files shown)


generate/src/contract.rs
```diff
@@ -77,6 +77,7 @@ pub(crate) fn expand_contract(args: &Args) -> Result<TokenStream> {
         }
 
         #[allow(dead_code)]
+        #[allow(clippy::too_many_arguments, clippy::type_complexity)]
```
bh2smith, line 80: I like it a lot!

bh2smith (approved):

PR #79, opened 2019-12-18
Yarn

nlordell:
:point_up: Yarn actually speeds up CI a lot, so definitely worth it.

### Test Plan

CI


bh2smith (approved):
Who knew that "using yarn instead of npm" was literally as easy as finding and replacing all occurrences of "npm" with "yarn"!

PR #82, opened 2019-12-19
Automatic Travis Deployment on Tags on Master Branch

nlordell:
This allows to just bump cargo versions and create a git tag and travis will handle the `cargo publish` for us.

### Test Plan

Deploy 0.2.0

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -68,3 +68,10 @@ script:
 
 before_cache:
   - rm -rf "$TRAVIS_HOME/.cargo/registry"
+
+deploy:
+  provider: script
+  script: bash ci/deploy.sh
+  on:
+    branch: master
+    tags: true
--- a/.travis.yml
+++ b/.travis.yml
@@ -73,5 +73,4 @@ deploy:
   provider: script
   script: bash ci/deploy.sh
   on:
-    branch: master
     tags: true
--- a/ci/deploy.sh
+++ b/ci/deploy.sh
@@ -1,0 +1,59 @@
+#!/bin/bash
+
+set -e
+
+token=$CARGO_TOKEN
+tag=$TRAVIS_TAG
+options=""
+while [[ $# -gt 0 ]]; do
+	case $1 in
+		--token) token="$2"; shift;;
+		--tag) tag="$2"; shift;;
+		--dry-run) options+="--dry-run ";;
+		-v|--verbose) options+="--verbose ";;
+		-h|--help) cat << EOF
+ci/deploy.sh
+Deploy the workspace to crates.io
+
+USAGE:
+    $0 [OPTIONS]
+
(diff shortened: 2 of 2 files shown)


.travis.yml
```diff
@@ -68,3 +68,10 @@ script:
 
 before_cache:
   - rm -rf "$TRAVIS_HOME/.cargo/registry"
+
+deploy:
+  provider: script
+  script: bash ci/deploy.sh
+  on:
+    branch: master
+    tags: true
```
fleupold: Are you sure this works?

From: https://docs.travis-ci.com/user/deployment/#conditional-releases-with-on

> tags: true: deployment is triggered if and only if $TRAVIS_TAG is set. [...] **This causes the branch condition to be ignored.**

ci/deploy.sh
```diff
@@ -0,0 +1,59 @@
...
+token=$CARGO_TOKEN
+tag=$TRAVIS_TAG
+options=""
+while [[ $# -gt 0 ]]; do
+	case $1 in
+		--token) token="$2"; shift;;
+		--tag) tag="$2"; shift;;
+		--dry-run) options+="--dry-run ";;
+		-v|--verbose) options+="--verbose ";;
+		-h|--help) cat << EOF
```
fleupold, line 14: Damn, this is some serious shell script magic.

<a href="https://gitme.me/image?url=https%3A%2F%2Fmedia1.giphy.com%2Fmedia%2F12NUbkX6p4xOO4%2Fgiphy.gif&token=magic" data-gitmeme-token="magic"></a>

ci/deploy.sh
```diff
@@ -0,0 +1,59 @@
...
+	fi
+}
+
+check_manifest_version common/Cargo.toml
+check_manifest_version generate/Cargo.toml
+check_manifest_version derive/Cargo.toml
+check_manifest_version Cargo.toml
+
+function cargo_publish {
+	(cd $1; cargo publish --token $CARGO_TOKEN $options)
```
fleupold, line 53: nice trick to avoid `cd -` at the end

.travis.yml
nlordell: THanks! I must have missed it. I made the `CARGO_TOKEN` only available on master so it should fail to publish on non-master branch.

ci/deploy.sh
nlordell: I copy-pasted it from the internetz

fleupold: Mind putting the link as a reference into the comments?

nlordell: > Mind putting the link as a reference into the comments?

https://stackoverflow.com/a/33826763

bh2smith (approved):
Fancy stuff.

PR #90, opened 2019-12-20
Test minimum supported rust version with CI

nlordell:
Also have a  note in the README on the minimum supported Rust version.


bh2smith (approved):
Love it!

PR #107, opened 2020-01-07
Initial Confirmation `ConfirmFuture` Implementation

nlordell:
This PR adds a `ConfirmFuture` implementation similar to confirmations in `web3`. It addresses some issues that we ran into regarding the confirmations implementation in the `web3` crate. Namely, it supports confirmations on Ganache, waiting for a transaction to be mined without extra confirmations on testnet and mainnet nodes, and polling fallback in cases where filters are not supported by the node (Infura over HTTPS for example). More information can be found here #104.

This is the first step in unifying the confirmation API as described in issue #95 and a first chunk of the PR started here #105 

Note that this PR just introduces the `ConfirmFuture` with unit tests and exports it but does not use it anywhere. That will be added in a future PR as it requires some refactoring of the `SendFuture`.

### Test Case

Many new unit tests!!!

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,6 +33,7 @@ ethcontract-common = { version = "0.2.0", path = "./common" }
 ethcontract-derive = { version = "0.2.0", path = "./derive" }
 ethsign = "0.7"
 futures = { version = "0.3", features = ["compat"] }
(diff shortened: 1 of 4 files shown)


nlordell:
Closing as this was superseded by #105

PR #111, opened 2020-01-08
[Easy] Use `Self` instead of type name

### LINKED ISSUE #110: Use `Self` Where Possible
We have a lot of builders, using `Self` instead of the actual builder type on the return types is not only more concise but less prone to copy-pasta bugs.

nlordell:
Prefer `Self` over the typename - this is a stylistic change that helps with the builder patterns and makes refactoring a little easier.

Closes #110 

On hold pending merge of #105

### DIFF
--- a/common/src/truffle.rs
+++ b/common/src/truffle.rs
@@ -32,7 +32,7 @@ pub struct Artifact {
 
 impl Artifact {
     /// Creates an empty artifact instance.
-    pub fn empty() -> Artifact {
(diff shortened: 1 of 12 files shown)


fleupold (approved):
Surprised there isn't a clippy rule for it.

🍬

nlordell:
It would be cool to have! It turns out to be super useful, especially with the builder pattern

PR #114, opened 2020-01-08
[Easy] Add Error Context to Smart Contract Bindings Generation

### LINKED ISSUE #106: Better Errors For `ethcontract::contract` Macro
It can be very difficult to debug issues where the contract JSON cannot be found, improve errors when locating the JSON file fails.

nlordell:
Closes #106

### DIFF
--- a/generate/src/contract/deployment.rs
+++ b/generate/src/contract/deployment.rs
@@ -1,6 +1,6 @@
 use crate::contract::{methods, Context};
 use crate::util;
-use anyhow::Result;
+use anyhow::{Context as _, Result};
 use ethcontract_common::truffle::abi::{Param, ParamType};
 use inflector::Inflector;
 use proc_macro2::{Literal, TokenStream};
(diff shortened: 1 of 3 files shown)


generate/src/contract/deployment.rs
```diff
@@ -1,6 +1,6 @@
 use crate::contract::{methods, Context};
 use crate::util;
-use anyhow::Result;
+use anyhow::{Context as _, Result};
```
bh2smith, line 3: What is the underscore being used for? I don't see it mentioned again later in the "Context"!

nlordell: To use trait methods in Rust, the trait must be imported into the local scope. If you want to use the trait methods only but not actually use the trait anywhere else, you can import it `as _` so that you can call methods defined by the trait without worrying about name collisions. Here we have a struct called `Context` as well so using this anonymous import avoids any compiler errors from from `Context` already being in use.

nlordell: The `Context` trait is needed for the `context` and `with_context` methods on the `anyhow::Result` type.

nlordell:
Re-opened to give Travis a nudge.

fleupold (approved):
<a href="https://gitme.me/image?url=https%3A%2F%2Fmedia1.giphy.com%2Fmedia%2FghCeBSsrXl3igHacZc%2Fgiphy.gif&token=improve" data-gitmeme-token="improve"></a>

PR #116, opened 2020-01-09
[Easy] Fail cargo build if lock file is not up to date

nlordell:
:point_up: :point_up_2:


nlordell:
Oops, this is a library and as such does not have a lock file!

PR #124, opened 2020-01-13
Now or Never

### LINKED ISSUE #120: Unit Tests Should Not Call Wait.
We should use something like [`now_or_never`](https://docs.rs/futures/0.3.1/futures/future/trait.FutureExt.html#method.now_or_never) instead.

nlordell:
Right now, tests use `futures::future::block_on` to wait for futures to resolve. This can be problematic if code causes tests to deadlock (because of a bug for example). Instead tests should use `FutureExt::now_or_never` which returns an `Option<Future::Output>` that is `Some(value)` if the future resolved immediately or `None` if it didn't. This asserts that test futures resolve immediately which is the case for all our tests.

We wrap `now_or_never` in an `FutureImmediateExt::immediate` helper method that additionally panics if the future did not resolve immediately for our tests.

Closes #120.

### Test Plan

CI

### DIFF
--- a/src/test/prelude.rs
+++ b/src/test/prelude.rs
@@ -2,20 +2,20 @@
 
 pub use crate::test::macros::*;
 pub use crate::test::transport::TestTransport;
+use futures::future::FutureExt;
 pub use serde_json::json;
 use std::future::Future;
 pub use web3::api::Web3;
 
-/// Temporary solution to work around the issue that async tests are not stable
-/// and the extra boiler plate of setting up an executor is inconvenient.
-/// TODO(nlordell): remove once async tests stablalize
-pub trait FutureWaitExt: Future {
+/// An extension future to assert that a future resolves immediately.
+pub trait FutureImmediateExt: Future {
     /// Block thread on a future completing.
-    fn wait(self) -> Self::Output;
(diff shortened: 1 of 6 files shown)


src/test/prelude.rs
```diff
@@ -2,20 +2,20 @@
...
+    fn immediate(self) -> Self::Output;
 }
 
-impl<F: Future + Sized> FutureWaitExt for F {
-    fn wait(self) -> Self::Output {
-        futures::executor::block_on(self)
+impl<F: Future + Sized> FutureImmediateExt for F {
+    fn immediate(self) -> Self::Output {
+        self.now_or_never()
+            .expect("future did not resolve immediately")
```
bh2smith, line 19: `Now` this is my kinda function! I `Never` could have lived without it!

bh2smith (approved):

e00E (approved):

fleupold (approved):
<a href="https://gitme.me/image?url=https%3A%2F%2Fmedia1.giphy.com%2Fmedia%2Ftn8zWeNYA73G0%2Fgiphy.gif&token=now" data-gitmeme-token="now"></a>

PR #125, opened 2020-01-13
No more manual pinning and `Unpin` implementations for futures.

### LINKED ISSUE #118: [Refactor] Use pin projections instead of relying on everything being `Unpin`

nlordell:
This PR removes `Web3Unpin` which was not really sound.

In re-reading the [`std::pin` documentation about structural pinning](https://doc.rust-lang.org/std/pin/index.html#projections-and-structural-pinning) it looks like a type can safely implement `Unpin` even if its fields are not all `Unpin`:
> You may also impl Unpin for Struct even if the type of field is not Unpin. What that type thinks about pinning is not relevant when no Pin<&mut Field> is ever created.

For a type to correctly implement `Unpin`, it requires that all fields that require structural pinning be `Unpin`:
> The struct must only be Unpin if all the structural fields are Unpin.

For futures:
> When implementing a Future combinator, you will usually need structural pinning for the nested futures, as you need to get pinned references to them to call poll. But if your combinator contains any other data that does not need to be pinned, you can make those fields not structural and hence freely access them with a mutable reference even when you just have Pin<&mut Self> (such as in your own poll implementation).

As such, when implementing a future combinator, the future can correctly implement `Unpin` if all of its nested futures are also `Unpin`. Since this is the case for `Compat01As03` future, which is the type for all the nested futures `ethcontract` builds on, all of `ethcontract` futures can also implement `Unpin`.

In order to ensure that the `Unpin` implementations are sound, we use `pin_project` crate that will add the correctly bounded `Unpin` implementations for all our structs as well as pin projections so the future implementations no longer need to worry about the complexities of pinning.

This PR closes #118 

### Test Plan

CI


src/contract/method.rs
```diff
@@ -131,19 +131,22 @@ impl<F> MethodFuture<F> {
     fn new(function: Function, inner: F) -> Self {
         MethodFuture { function, inner }
     }
+
+    /// Get a pinned projection to the inner future.
+    fn inner(self: Pin<&mut Self>) -> Pin<&mut F> {
+        unsafe { self.map_unchecked_mut(|f| &mut f.inner) }
```
fleupold: Oh my, do you really want to add unsafe code to your library? Would it be possible to wrap inner into a `Pin<Box<F>>` or use a library like [pin_project](https://docs.rs/pin-project/0.4.6/pin_project/index.html)?

src/transaction.rs
```diff
@@ -582,6 +582,13 @@ impl<T: Transport> SendFuture<T> {
     }
 }
 
+// NOTE: This is safe since `SendFuture` only requires pinned `SendState` and
```
fleupold: Can you explain how the alternative with trait bounds would work?

fleupold (commented):
This is pretty complex stuff. I watched https://www.youtube.com/watch?v=DkMwYxfSYNQ to get a better understanding of Pinning (they suggesting using Pin<Box> or the pin_project crate instead of writing unsafe code.

> Since this is the case for Compat01As03 future, which is the type for all the nested futures ethcontract builds on, all of ethcontract futures can also implement Unpin.

Why does MethodFuture and MaybeReady then need the unsafe code and cannot be constrained to only having Unpin inner futures?

src/contract/method.rs
nlordell: Here be dragons... This is what pin_project generates for us anyway.

nlordell: I can add some comments as to why this is ok to do or require that these take Unpin futures.

src/transaction.rs
nlordell: https://github.com/gnosis/ethcontract-rs/blob/080d884ec904d8574b4c2e7628860daaa5062da0/src/contract/deploy.rs#L72

This is an example. We are basically declaring that only the inner future needs to be `Unpin` for the outer future to be it as well since it’s the only thing that we need pinned references to.

nlordell:
Yeah, I’m not 100% confident with this whole pinning concept either, but in re-reading the pin module docs I came to the conclusion that what we were doing before was slightly incorrect.

MaybeReady and MethodFuture are only ever used with `Unpin` futures it doesn’t technically need to work with `!Unpin` futures. I can roll back that change if you are uncomfortable with the unsafe code.

src/transaction.rs
fleupold: And why can't we do that inside SendFuture? How would that leak the SendState type?

src/contract/method.rs
fleupold: I think there is still a difference if you are using a library that is using unsafe code or if you are writing it yourself. E.g. at some point someone might go ahead and declare MethodFuture as Unpin, which as far as I understand would be unsafe. However with Pin<Box> that would be ok and pin_project has some other smart guarantees (e.g. that it doesn't let you incorrectly override the destructor).

Generally I'd defer to well tested and widely adopted libraries when relying on unsafe code and not encourage it to be used directly in our library.

nlordell: `impl Unpin for MethodFuture` is incorrect but never unsafe in itself, but it would require `unsafe` in order to pin the inner future for polling. That unsafe code would be wrong and potentially cause undefined behaviour.

That is a fair point, the smart guarantees of `pin_project` are definitely worth it to not have to roll these unsafe projections by hand.

src/transaction.rs
nlordell: Because `SendState` is a private type and would appear in a public interface (here the public interface is the trait bounds of a public trait on a public type). Specifically the compiler error is:
```
error[E0446]: private type `transaction::confirm::ConfirmState<T>` in public interface
   --> src/transaction/confirm.rs:145:1
    |
98  |   enum ConfirmState<T: Transport> {
    |   - `transaction::confirm::ConfirmState<T>` declared as private
...
145 | / impl<T> Unpin for ConfirmFuture<T>
146 | | where
147 | |     T: Transport,
148 | |     ConfirmState<T>: Unpin,
149 | | {
150 | | }
    | |_^ can't leak private type
```

src/contract/deploy.rs
```diff
@@ -58,30 +58,34 @@ where
...
         DeployedFuture {
-            args: Some((web3.into(), context)),
+            args: Some((web3, context)),
             network_id: net.version().compat(),
             _instance: PhantomData,
         }
     }
 }
 
+impl<T, I> Unpin for DeployedFuture<T, I>
```
fleupold: Shouldn't this be autogenerated for us if CompatCallFuture<T, String> is Unpin? 

Is it because the T we are using are not Unpin? If that's the case, are we sure web3 doesn't use give out `Pin<T>` internally which could make it invalid for us to more the DeployedFuture Unpin?

src/transaction.rs
fleupold: This would not be necessary if we used pin_project as they would handle the unpinning for us, wouldn't it?

fleupold (commented):
I like that the unsafe code is gone. I'm still a little bit concerned about marking the composed types as unpin (how can we guarantee that the code we call doesn't give out Pins to the !Unpin types?) and would like to see if using pin_project would prevent us from using the assert_unpin macro.

src/transaction.rs
nlordell: It is still a problem with `pin_project` as the generated code also leaks this type. They have a work around to include the implementation in a function which makes the compiler generate a warning instead of an error but then it doesn't play nice with `cargo doc` as generated types types show up and look like this:

image

Note that ` __ConfirmFuture<'pin, T>` can't be clicked on so you can't really inspect or debug why a type is not `Unpin`.

This is an unfortunate downside to the `pin_project`.

src/contract/deploy.rs
nlordell: I think the idea here is that since we never pin `Web3<T>` it is not our responsibility to uphold any pinning guarantees. If `Web3<T>` would pin a `T: !Unpin` internally for some operation, then it would require to write `unsafe` code, and it would be responsible for upholding the pinning guarantee.

src/transaction.rs
nlordell: If you're curious this is what it looks like: https://github.com/taiki-e/pin-project/blob/master/examples/struct-default-expanded.rs#L55-L90

nlordell:
@fleupold Since Github GUI doesn't let me respond to https://github.com/gnosis/ethcontract-rs/pull/125#discussion_r366772369 I will answer here:
> This would not be necessary if we used pin_project as they would handle the unpinning for us, wouldn't it?

Correct, it is now gone.

nlordell:
We are now completely free of manual `Unpin` implementations as well as `Pin::new` and `Pin::new_unchecked` calls! So we don't really have to worry about pinning at all in our code :).

src/contract/deploy.rs
nlordell: And yes, `CompatCallFuture` wasn't auto-implementing `Unpin` because of `T` in `Web3<T>`.

fleupold: >... it would require to write unsafe code, and it would be responsible for upholding the pinning guarantee.

My understanding is that since they mark T !Unpin, they could e.g. be giving out a Pin<T> via unsafe code without risking any undefined behavior. However, if we then marked the composite struct Unpin, we could actually end up moving T which might have been given out by the lower layer and thus cause undefined behaviour.

But I'm not sure this logic makes sense, since we could already move the T from a plain mutable instance without any Pin/Unpin.

src/contract/deploy.rs
```diff
@@ -72,15 +75,15 @@ where
 {
     type Output = Result<I, DeployError>;
 
-    fn poll(self: Pin<&mut Self>, cx: &mut Context) -> Poll<Self::Output> {
-        let unpinned = self.get_mut();
-        Pin::new(&mut unpinned.network_id)
+    fn poll(mut self: Pin<&mut Self>, cx: &mut Context) -> Poll<Self::Output> {
+        self.as_mut()
```
fleupold, line 79: Interesting that we are now requiring a mutable instance of self.

nit: `as_mut` could be avoided if web3, context get queried outside of the map closure.

src/contract/deploy.rs
nlordell: Its not that `T: !Unpin` but more that `T: ?Unpin` in that it can be `Unpin` since there are no bounds on T (besides being `Transport + Clone`). And, `Web3` is defined something like this: `struct Web3<T: Transport> { transport: T }` so our futures can't auto implement `Unpin` because `Web3<T>` can't because `T` may be unpin (as there is no trait bound).

In practice, none of the transports are `!Unpin`, the following compiles:
```
fn main() {
    use web3::transports::*;
    fn assert_unpin<T: Unpin>() {}
    assert_unpin::<Http>();
    assert_unpin::<WebSocket>();
    assert_unpin::<Ipc>();
    assert_unpin::<Batch<Http>>();
}
```

That being said, suppose that there is an transport that is `T: !Unpin` (which is totally possible) our future always receives an owned `Web3<T>`, which is not pinned and as such, our instance of `Web3` is completely free to move around. If there were for example a method `Web3::do_something(self: Pin<&mut Self)`, then that would require the future to pin the `Web3` instance. As soon as the `Web3` instance is pinned (which would be an `unsafe` thing since `T: !Unpin` in our example), then it would be the future's responsibility once the `Web3` is pinned, to ensure that it never moves.

If a transport were self referential (for example it kept some shared buffers or something) then it most probably would need to be constructed in a way such that it is always pinned. And you would only ever get a `Pin<Box<T>>` to it. In that case, the `Pin<Box<T>>` would be safe to move around (as it would effectively just be copying a pointer and not the data being pointed to).

From everything I understood, `T: !Unpin` can move freely until pinned. This is why, in the `futures` crate you have something like this:
```
pub struct Ready<T>(Option<T>);
impl<T> Unpin for Ready<T> {}
```
(https://docs.rs/futures-util/0.3.1/src/futures_util/future/ready.rs.html#8-10)

Note that if we ever required a pinned `Web3<T>`, then the `pin_project` macro would automatically add `where Web3<T>: Unpin` as a constraint to the future's `Unpin` implementation. In that case, where `Web3<T>` requires pinning to be used, then indeed the future would only be `Unpin` if `Web3<T>: Unpin` which is equivalent to `T: Unpin`.

fleupold (approved):
<a href="https://gitme.me/image?url=https%3A%2F%2Fmedia3.giphy.com%2Fmedia%2F7p0qZOxUe5cIM%2Fgiphy.gif&token=long_way" data-gitmeme-token="long_way"></a>

src/contract/deploy.rs
nlordell: This is another big pro to using `pin_project` as we don't have to remember to add these trait bounds in the case we need additional pinning on fields in our struct (most commonly this will happen if another future was added to the future combinator).

Also, since using `pin_project`, we no longer `impl Unpin` for anything (we defer that to the crate as it does it programmatically and makes sure that we don't forger to update anything). Check it out:
```
$ rg Unpin
$ rg Pin::new
$ 
```

src/contract/deploy.rs
nlordell: Yeah, that is since `fn as_mut(&mut Pin<P>) -> Pin<&mut <P as Deref>::Target>`. Its so the borrow checker makes sure that from a single `Pin<P>` you can't get multiple `Pin<&mut <P as Deref>::Target>` which would mean that you have multiple mutable references to the same thing :fearful:

nlordell: The problem if we take it out of the closure, is we `.take()` if from the `Option`, so if the future is not resolved the first time its polled, then it will panic the second time.

**edit**: ~~We can `project` to avoid the `as_mut`.~~ that won't work because then the `network_id` future gets moved out of the projection and we would need an `as_mut` there.

fleupold: Ah, you are right. Let's keep it as is then.

PR #127, opened 2020-01-16
Manual deployment address

### LINKED ISSUE #99: Option to generate `Contract::deployed` regardless of whether there are networks
This makes things like running `cargo test` easier.

nlordell:
Add ability to manually specify deployment addresses. This closes #99 and #100 .

This PR also adds unit tests in the code generation and derive libraries - so unit tests can be slowly added for code generation as future modification come in.

### Test Plan

Unit tests and new example.

### DIFF
--- a/derive/src/lib.rs
+++ b/derive/src/lib.rs
@@ -5,35 +5,74 @@
 
 extern crate proc_macro;
 
+mod spanned;
+
+use crate::spanned::{ParseInner, Spanned};
 use ethcontract_generate::{parse_address, Address, Builder};
 use proc_macro::TokenStream;
+use proc_macro2::Span;
+use std::collections::HashSet;
 use syn::ext::IdentExt;
 use syn::parse::{Error as ParseError, Parse, ParseStream, Result as ParseResult};
 use syn::{braced, parse_macro_input, Error as SynError, Ident, LitInt, LitStr, Token};
 
-/// Proc macro to generate type-safe bindings to a contract. See
-/// [`ethcontract`](ethcontract) module level documentation for more information.
+/// Proc macro to generate type-safe bindings to a contract. This macro accepts
+/// a path to a Truffle artifact JSON file. Note that this path is rooted in
+/// the crate's root `CARGO_MANIFEST_DIR`.
+///
+/// ```ignore
+/// ethcontract::contract!("build/contracts/MyContract.json");
+/// ```
+///
+/// Currently the proc macro accepts additional parameters to configure some
+/// aspects of the code generation. Specifically it accepts:
+/// - `crate`: The name of the `ethcontract` crate. This is useful if the crate
+///   was renamed in the `Cargo.toml` for whatever reason.
+/// - `deployments`: A list of additional addresses of deployed contract for
+///   specified network IDs. This mapping allows `MyContract::deployed` to work
+///   for networks that are not included in the Truffle artifact's `networks`
+///   property. Note that deployments defined this way **take precedence** over
+///   the ones defined in the Truffle artifact. This parameter is intended to be
+///   used to manually specify contract addresses for test environments, be it
+///   testnet addresses that may defer from the originally published artifact or
+///   deterministic contract addresses on local development nodes.
+///
+/// ```ignore
+/// ethcontract::contract!(
+///     "build/contracts/MyContract.json",
+///     crate = ethcontract_rename,
+///     deployments {
+///         4 => "0x000102030405060708090a0b0c0d0e0f10111213"
+///         5777 => "0x0123456789012345678901234567890123456789"
+///     },
+/// );
+/// ```
+///
+/// See [`ethcontract`](ethcontract) module level documentation for additional
+/// information.
 #[proc_macro]
 pub fn contract(input: TokenStream) -> TokenStream {
-    let args = parse_macro_input!(input as ContractArgs);
+    let args = parse_macro_input!(input as Spanned<ContractArgs>);
 
-    let artifact_span = args.artifact_path.span();
-    args.into_builder()
+    let span = args.span();
+    args.into_inner()
+        .into_builder()
         .generate()
         .map(|bindings| bindings.into_tokens())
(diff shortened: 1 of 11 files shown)


derive/src/lib.rs
```diff
@@ -27,48 +28,167 @@ pub fn contract(input: proc_macro::TokenStream) -> TokenStream {
...
-//   experience.
+impl ContractArgs {
+    fn into_builder(self) -> Builder {
+        let mut builder = Builder::new(&self.artifact_path.value());
+        for parameter in self.parameters.into_iter() {
+            builder = match parameter {
+                Parameter::Crate(name) => builder.with_runtime_crate_name(name),
+                Parameter::Deployments(deployments) => {
+                    deployments.into_iter().fold(builder, |builder, d| {
+                        builder.add_deployment(d.network_id, d.address)
```
fleupold, line 81: What should we do in case there already exists a deployment in the artifact? Right now we override. Should we warn or even assert that it doesn't yet exist?

generate/src/contract.rs
```diff
@@ -61,11 +63,24 @@ impl Context {
             artifact,
             runtime_crate,
             contract_name,
```
fleupold, line 66: Do we still need these two if we have artifact and args directly? Generally I'd slightly prefer to keep the way it is right now and only take the fields we actually need. In this case storing deployments instead of args should be enough, right?

derive/src/lib.rs
```diff
@@ -27,48 +28,167 @@ pub fn contract(input: proc_macro::TokenStream) -> TokenStream {
...
+                4 => "0x0123456789012345678901234567890123456789",
+            },
+        );
+        assert_eq!(
+            args.parameters,
+            &[
+                Parameter::Crate("foobar".into()),
+                Parameter::Deployments(vec![
+                    deployment(1, "0x000102030405060708090a0b0c0d0e0f10111213"),
+                    deployment(4, "0x0123456789012345678901234567890123456789"),
```
bh2smith: Are you running a deterministic ganache?

bh2smith (approved):
I gotta say, this was a lot to look at... Not sure if I would be able to catch any big problems if there were any.

generate/src/lib.rs
```diff
@@ -60,15 +65,46 @@ impl Builder {
...
     }
 
+    /// Manually adds specifies the deployed address as a string of a contract
+    /// for a given network. See `Builder::add_deployment` for more information.
+    ///
+    /// # Panics
+    ///
+    /// This method panics if the specified address string is invalid. See
+    /// `parse_address` for more information on the address string format.
+    pub fn add_deployment_str<S>(self, network_id: u32, address: S) -> Self
```
fleupold, line 95: Are we using this?

fleupold (approved):
<a href="https://gitme.me/image?url=https%3A%2F%2Fmedia3.giphy.com%2Fmedia%2FjKaFXbKyZFja0%2Fgiphy.gif&token=gif" data-gitmeme-token="gif"></a>

derive/src/lib.rs
nlordell: No, this is just a unit test to make sure that the proc macro parses the deployments correctly.

generate/src/lib.rs
nlordell: No, but it makes it easier to use when using `ethcontract_generate` in `build.rs`. I wrote this with `dex-services` in mind, to avoid having to write the `parse_address` code there.

The idea is that `dex-services` will call this with the deterministic contract addresses so that there is no more requirement to run `truffle migrate` before building.

generate/src/contract.rs
nlordell: My motivation here was that as `Args` grows, we don't have to keep adding new fields to `Context` - but I agree, it will probably be cleaner than just keeping the whole `Args` around.

The `runtime_crate` and `contract_name` come from the `Args` but have been transformed to `Ident`s so they can be `quote!`-ed.

derive/src/lib.rs
nlordell: That's a good point. I will look into propagating compiler warnings so they see this. The idea here would be if you do this:
```
ethcontract::contract!(
    "path/to/artifact",
    deployments {
        1 => "0x...",
        2 => "0x...",
        1 => "0x...", // should be a compiler warning on this line
    }
);
```

nlordell: Currently, Rust doesn't support warnings from proc_macros in stable (there is support on nightly, but I don't think it will be stabilized anytime soon). This means its not possible to warn when a manually specified contract address collides with a JSON artifact one.

I think that for now, overwriting the artifact's network configuration is the correct behaviour given the limitations. The idea of manually specified deployments is to be able to:
- provide deterministic addresses for testing scenarios with development nodes (like what we want for `dex-services` with Ganache tests)
- provide a way to override addresses from a truffle artifact (for example if we are generating bindings to a contract that has its artifact published on NPM but have our own deployment to the contract that we want to use instead).

The explicitness of the `deployments { ... }` parameter in the macro invocation makes be not feel too bad about it not being an error.

I did add a hard compile error in the case where a deployment gets specified more than once in the proc macro invocation:
```
error: duplicate network ID in `ethcontract::contract!` macro invocation
 --> examples/deployments.rs:9:9
  |
9 |         5777 => "0x0123456789012345678901234567890123456789",
  |         ^^^^
```

I also added a some more documentation about it here: https://github.com/gnosis/ethcontract-rs/blob/manual_deployment_address/derive/src/lib.rs#L31-L38

@fleupold Let me know if you are OK with this.

nlordell: One idea I just had was to slightly modify the syntax to be like this:
```
5777! => "0x..." // override network ID 5777
5777 => "0x..." // error if already exists in contract.
```

If you think that is a good alternative, I'll open a new issue and do it in a separate PR.

fleupold: I think that’s a very neat idea. Wondering if it should only be an error if the address in the contract is different?

nlordell: Created #130

PR #140, opened 2020-01-21
Move transaction futures into their own modules

nlordell:
The transaction module was over 1000 lines and was getting hard to navigate.

### DIFF
--- a/src/transaction.rs
+++ b/src/transaction.rs
@@ -1,26 +1,19 @@
 //! Implementation for setting up, signing, estimating gas and sending
 //! transactions on the Ethereum network.
 
+pub mod build;
 pub mod confirm;
+pub mod estimate_gas;
+pub mod send;
 
-use crate::errors::ExecutionError;
-use crate::future::{CompatCallFuture, MaybeReady};
-use crate::sign::TransactionData;
-use crate::transaction::confirm::{ConfirmFuture, ConfirmParams};
+use crate::transaction::build::BuildFuture;
+use crate::transaction::confirm::ConfirmParams;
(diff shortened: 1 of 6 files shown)


fleupold (approved):
<a href="https://gitme.me/image?url=https%3A%2F%2Fmedia1.giphy.com%2Fmedia%2FxT5LMPcAaKKiOzwGsw%2Fgiphy.gif&token=move" data-gitmeme-token="move"></a>

src/transaction.rs
```diff
@@ -70,6 +63,37 @@ impl Default for ResolveCondition {
...
+pub enum Transaction {
+    /// A structured transaction request to be signed locally by the node.
+    Request(TransactionRequest),
+    /// A signed raw transaction request.
+    Raw(Bytes),
+}
+
+impl Transaction {
+    /// Unwraps the transaction into a transaction request, returning None if the
+    /// transaction is a raw transaction.
```
bh2smith, line 79: Don't all successful transactions return a receipt of some type (regardless of whether we were expecting a return value).

bh2smith (approved):
Once again, this all looks really great. I am impressed by all the dedication to doc-strings, but I am unsure if I can provide a proper review here. There is so much going on.

src/transaction.rs
nlordell: This is for sending transactions, they can either be a transaction object (which is unsigned and requires the node to sign transactions locally) or a raw transactions (a bunch of bytes, that are signed). This utility method just returns `Some(TransactionRequest)` if it is the former, and `None` if it is the latter.

PR #145, opened 2020-01-24
Gate `ethcontract-derive` dependency on feature

### LINKED ISSUE #91: Gate `ethcontract-derive` Dependency on Feature
This will allow for dependent crates to not build `ethcontract-derive` when using the `ethcontract-generate` API.

e00E:
This will allow for dependent crates to not build ethcontract-derive when
using the ethcontract-generate API.

fixes #91

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,7 +17,7 @@ to Ethereum smart contracts.
 name = "ethcontract"
 
 [features]
-default = []
+default = ["ethcontract-derive"]
 samples = []
 
 [workspace]
@@ -30,7 +30,7 @@ members = [
 
 [dependencies]
 ethcontract-common = { version = "0.2.0", path = "./common" }
-ethcontract-derive = { version = "0.2.0", path = "./derive" }
+ethcontract-derive = { version = "0.2.0", path = "./derive", optional = true}
 ethsign = "0.7"
 futures = { version = "0.3", features = ["compat"] }
 futures-timer = "2.0"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,7 +17,8 @@ to Ethereum smart contracts.
 name = "ethcontract"
 
 [features]
-default = ["ethcontract-derive"]
+default = ["derive"]
+derive = ["ethcontract-derive"]
 samples = []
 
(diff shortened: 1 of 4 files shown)


bh2smith (approved):
Interesting. Do we have a concrete scenario when one could benefit from such a change?

e00E:
I implemented this because there was an issue for it.
Generally, it is helpful to put parts of a crate into optional features to allow users of the create to only compile what they need. This reduces compile times and the total number of dependencies.

e00E:
Can I get travis to try again? I am not sure if the CI errror is related to my change because Felix mentioned another travis problem in Slack.

generate/README.md
```diff
@@ -17,6 +17,8 @@ generator:
 ```toml
 [dependencies]
 ethcontract = "..."
```
nlordell:
```suggestion
ethcontract = { version = "...", no-default-features = true }
```

Then we no longer need the comment below as well.

examples/generate/Cargo.toml
```diff
@@ -6,7 +6,7 @@ authors = ["Nicholas Rodrigues Lordello <nlordell@gmail.com>"]
 edition = "2018"
 
 [dependencies]
-ethcontract = { path = "../.." }
+ethcontract = { path = "../..", no-default-features = true }
```
nlordell, line 9: :metal:

Cargo.toml
```diff
@@ -17,7 +17,7 @@ to Ethereum smart contracts.
 name = "ethcontract"
 
 [features]
-default = []
+default = ["ethcontract-derive"]
```
nlordell: nit: can we name the future `derive`?

nlordell (commented):
Just a few minor comments

examples/generate/Cargo.toml
e00E: I am interpreting this as a sign of approval :) .

nlordell: It was a great catch!

nlordell (approved):
Looks great!

I think having it be a default additive feature is definitely the way to go!

PR #147, opened 2020-01-27
Update web3 to v0.10.0

### LINKED ISSUE #139: Update to `web3 v0.10.0`
Big update, this also allows us to resolve:
- [x] #8 
- [x] ~~#14~~ Signing was removed from `v0.10.0` so it will have to wait until the next version.
- [x] ~~#138~~ Gas estimation in `web3` does not work for contract deployment.

e00E:


nlordell:
Looks good to me! I would, however, I don't think that we can close #139 just yet with this, as there are 3 other related cases.

Also, can we hold off merging this until `0.3.0` so we don't have to deal with version issues in `dex-services`?.

e00E:
Doesn't matter to me whether the issue stays open. I do think its cleaner to have separate PRs for the *update* itself and the other issues.

We can definitely hold off merging.

nlordell:
And we definitely want multiple PRs

nlordell:
@e00E Can you rebase this? Preferably also bump to web3 0.10.0 :)

nlordell (approved):

nlordell:
Also, this can be merged once that's done. Since we removed `graph` dependency from services, we no longer have any motivation to keep the web3 version pinned back.

PR #150, opened 2020-01-27
[Easy] Correctly specify no 'default-features' for generate example

nlordell:
Looks like the correct way to disable default features in a `Cargo.toml` is `default-features = false`. This fixes that.

### Test Plan

1) See the build warning here: https://travis-ci.org/gnosis/ethcontract-rs/jobs/642532349#L271
  ```
warning: /home/travis/build/gnosis/ethcontract-rs/examples/generate/Cargo.toml: unused manifest key: dependencies.ethcontract.no-default-features
  ```
  Or checkout https://github.com/gnosis/ethcontract-rs/commit/0607e87002d2299d93264e1219e09774f9f7c2a3 and run a `cargo build`.

2) Check that this is no longer the case.

### DIFF
--- a/examples/generate/Cargo.toml
+++ b/examples/generate/Cargo.toml
@@ -6,7 +6,7 @@ authors = ["Nicholas Rodrigues Lordello <nlordell@gmail.com>"]
 edition = "2018"
 
 [dependencies]
-ethcontract = { path = "../..", no-default-features = true }
+ethcontract = { path = "../..", default-features = false }
 futures = "0.3"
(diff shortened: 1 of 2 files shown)


e00E:
I see, `--no-default-features` can be passed to `cargo build`, but not in the toml.

e00E (approved):

nlordell:
It also just leaves a warning which is super easy to miss. I wish there were a way to fail the build when there are unknown manifest keys but I couldn't find anything.

PR #159, opened 2020-01-29
Fix infinite loops in replace_in_place

e00E:
If from and to are both empty then the while loop runs forever because
self.find always returns 0.

If from and to are equal then the while loop runs forever because it
will keep finding and replacing the first result.


nlordell (approved):
Love it! Thanks for the unit tests too :metal:

PR #162, opened 2020-01-30
Use str::replace instead of replace_in_place

### LINKED ISSUE #148: Not sure if replace_all_in_place is safe

e00E:
In order to avoid unsafe.

fixes #148


e00E:
Should I move `to_fixed_hex` into `bytecode.rs` too?

common/src/bytecode.rs
```diff
@@ -72,10 +72,10 @@ impl Bytecode {
         //   `LinkedContract` contract for and example of how it looks like
         let placeholder = format!("__{:_<38}", name);
         let address = address.to_fixed_hex();
-        let matched = self.0.replace_all_in_place(&placeholder, &address);
-        if !matched {
+        if self.0.find(&placeholder).is_none() {
             return Err(LinkError::NotFound(name.to_string()));
         }
+        self.0 = self.0.replace(&placeholder, &address);
```
nlordell, line 77: Can we add a unit test similar to https://github.com/gnosis/ethcontract-rs/pull/162/files#diff-3da5111806a318876c8d54674237ec40L69 that checks that multiple occurrences of the same library get replaced?

nlordell (commented):
@e00E, yeah lets move the `to_fixed_hash` method into `bytecode` module. Feel free to make it a function and not a extension trait method too.

common/src/bytecode.rs
e00E: I added a test that makes sure that str::replace works as expected. This has less boilerplate than creating a whole ByteCode from scratch but let me know if that is what you had in mind.

common/src/bytecode.rs
```diff
@@ -207,4 +210,34 @@ mod tests {
...
+            ),
+        ] {
+            let value: Address = value[2..].parse().unwrap();
+            assert_eq!(super::to_fixed_hex(&value), *expected);
+        }
+    }
+
+    #[test]
+    fn string_replacement_replaces_multiple_occurrences() {
+        // This is important for Bytecode::link.
```
nlordell: I would prefer if this method test `Bytecode::Link` directly. And maybe a second test to verify the branch where the library is not needed?

common/src/bytecode.rs
```diff
@@ -71,11 +70,11 @@ impl Bytecode {
         //   '__$name__..__' with the library address; see generated bytecode for
         //   `LinkedContract` contract for and example of how it looks like
         let placeholder = format!("__{:_<38}", name);
-        let address = address.to_fixed_hex();
-        let matched = self.0.replace_all_in_place(&placeholder, &address);
-        if !matched {
+        let address = to_fixed_hex(&address);
+        if self.0.find(&placeholder).is_none() {
```
nlordell: nit: `!self.0.contains(&placeholder)`

nlordell (approved):
Looks great!

common/src/bytecode.rs
```diff
@@ -207,4 +210,63 @@ mod tests {
...
+        let placeholder = format!("__{:_<38}", "name0");
+        let mut bytecode = Bytecode::from_hex_str(format!(
+            "0x61{}{}61{}",
+            placeholder, placeholder, placeholder
+        ))
+        .unwrap();
+        // name does not match
+        match bytecode.link("name1", address) {
+            Err(LinkError::NotFound(_)) => (),
+            _ => assert!(false, "should fail with not found error"),
```
nlordell: Its failing because of a clippy lint:
```suggestion
            _ => panic!("should fail with not found error"),
```

PR #163, opened 2020-01-30
Fix ci for beta and nightly rust

e00E:
CI started failing on beta and nightly channels with the clippy warning
option-as-ref-deref with this suggested fix.


e00E:
This fix is needed for beta and nightly but not allowed for stable. How do we make CI happy?

nlordell:
You have a few options:
- Bump minimum Rust version to 1.40
- `#[allow(clippy::option-as-ref-deref)]`.
- Only enable clippy lints on stable (not on 1.39, beta or nightly).
- Add a shim (`OptionExt` extension trait only for rust version <1.40) (personally not a fan)

nlordell:
Personally I suggest bumping the minimum rust version to 1.40. You will need to update the travis file and the README as well.

e00E:
I have removed the explanatory part of the minimum rust version because there is more than one reason now and it isn't useful to track why this is the minimum version if we are using several normal rust features.

nlordell (approved):
Ship it!

PR #175, opened 2020-02-06
[Easy] Fix nightly redundant import lint

nlordell:
The nightly build was catching this lint, lets fix it nice and early.

Here is the lint in question: https://travis-ci.org/gnosis/ethcontract-rs/jobs/646874470#L1329

### Test Plan

CI + check that the nightly build is no longer failing.


bh2smith (approved):
iz good

PR #178, opened 2020-02-07
Test deploy

### LINKED ISSUE #2: Add unit tests for contract deployment.

e00E:
fixes #2

### DIFF
--- a/src/contract/deploy.rs
+++ b/src/contract/deploy.rs
@@ -247,12 +247,6 @@ mod tests {
     }
 
     #[test]
-    fn deploy() {
-        // TODO(nlordell): implement this test - there is an open issue for this
-        //   on github
-    }
-
-    #[test]
     fn deploy_fails_on_empty_bytecode() {
         let transport = TestTransport::new();
         let web3 = Web3::new(transport.clone());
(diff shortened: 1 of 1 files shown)


e00E:
@nlordell Is this roughly what you had in mind? Asking before I add more tests.

src/contract/deploy.rs
```diff
@@ -264,4 +258,63 @@ mod tests {
...
+            .unwrap();
+
+        transport.assert_request("eth_accounts", &[]);
+        transport.assert_request(
+            "eth_sendTransaction",
+            &[json!({"data": bytecode, "from": account})],
+        );
+        transport.assert_request("eth_blockNumber", &[]);
+        transport.assert_request("eth_getTransactionReceipt", &[json!(transaction_hash)]);
+        transport.assert_no_more_requests();
```
nlordell, line 318: Maybe we can assert that the `Instance` result address is matches what was returned from the tx receipt.

nlordell (commented):
That was exactly what I had in mind. What other `deploy` tests did you have in mind?

src/contract/deploy.rs
```diff
@@ -264,4 +258,63 @@ mod tests {
         assert_eq!(error.to_string(), DeployError::EmptyBytecode.to_string());
         transport.assert_no_more_requests();
     }
+
+    #[test]
+    fn deploy() {
+        let mut transport = TestTransport::new();
+        let web3 = Web3::new(transport.clone());
+
+        let mut artifact = Artifact::empty();
```
nlordell, line 267: nit:
```
let artifact = Artifact {
    bytecode: Bytecode::from_hex_str(bytecode).unwrap(),
    ..Artifact::empty()
};
```

src/contract/deploy.rs
```diff
@@ -264,4 +258,63 @@ mod tests {
...
+        transport.add_response(json!(transaction_hash));
+
+        // eth_blockNumber
+        transport.add_response(json!("0x0"));
+
+        // eth_getTransactionReceipt
+        let contract_address = addr!("0x9876543210987654321098765432109876543210");
+        let block_hash =
+            hash!("0x4242424242424242424242424242424242424242424242424242424242424242");
+        let mut logs_bloom = "0x".to_string();
```
nlordell, line 288:
```suggestion
        let logs_bloom = H2048::repeat_byte(0);
```
https://docs.rs/web3/0.10.0/web3/types/struct.H2048.html#method.repeat_byte

nlordell: This can probably be fixed in other locations in the code.

e00E:
> That was exactly what I had in mind. What other `deploy` tests did you have in mind?

The ones listed in #2 .

nlordell:
@e00E Oh yeah, I forgot about my list. Sure, note that I think there might be a bug with transaction signing for transactions that deploy contracts.

Take a look here for how RLP encoding should work for transaction where the `to` address is `None`:
https://github.com/tomusdrw/rust-web3/blob/ab762e0f48c527a19952c3eb776fc7a7fb14f52e/src/api/accounts.rs#L267-L312

Opened issue #181

PR #179, opened 2020-02-07
More convenient TestTransport

e00E:
This changes `TestTransport` to have one `add_request_response` method instead of separate `assert_request` and `add_response`.
In my opinion this is nicer because you have the request and response right next to each other in tests and not first all responses and then all requests.

I only changed `TestTransport` and one test as an example so it doesn't compile yet. If you @nlordell  think its a good change I can apply it to all tests.

Other minor changes:
* created structs instead of the tuples
* turned the future error in `send` into a panic because I wasn't sure why it wasn't a panic in the first place

### DIFF
--- a/src/transport.rs
+++ b/src/transport.rs
@@ -107,15 +107,20 @@ mod tests {
         let mut transport = TestTransport::new();
         let dyn_transport = DynTransport::new(transport.clone());
 
-        // assert that the underlying transport prepares the request.
+        // assert that the underlying transport prepares the request and returns
+        // the response.
+        transport.add_request_response("test".to_string(), vec![json!(28)], json!(true));
         let (id, call) = dyn_transport.prepare("test", vec![json!(28)]);
-        transport.assert_request("test", &[json!(28)]);
         transport.assert_no_more_requests();
 
-        // assert that the underlying transport returns the response.
-        transport.add_response(json!(true));
         let response = dyn_transport.send(id, call).wait().expect("success");
         assert_eq!(response, json!(true));
+
+    #[test]
+    #[should_panic]
+    fn dyn_transport_panics() {
+        let mut transport = TestTransport::new();
+        let dyn_transport = DynTransport::new(transport.clone());
 
         // assert that the transport layer gets propagated - it errors here since
         // we did not provide the test transport with a response
@@ -124,7 +129,5 @@ mod tests {
             .wait()
             .err()
             .expect("failed");
-        transport.assert_request("test", &[json!(42)]);
-        transport.assert_no_more_requests();
     }
 }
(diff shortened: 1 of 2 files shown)


src/transport.rs
```diff
@@ -107,24 +107,27 @@ mod tests {
         let mut transport = TestTransport::new();
         let dyn_transport = DynTransport::new(transport.clone());
 
-        // assert that the underlying transport prepares the request.
+        // assert that the underlying transport prepares the request and returns
+        // the response.
+        transport.add_request_response("test".to_string(), vec![json!(28)], json!(true));
```
nlordell, line 112: My only issue with this is that this function call is a little hard to parse at first glance. Maybe `Request` and `Response` tuple structs would be nice? Not sure if that would be better or worse though...

Maybe you can try some other syntax to make it look a bit more legible. Another thing is the `.to_string()` on every invocation is tedious and can be avoided by taking any `Into<String>`. 

Another ergonomic improvement would be to take a `Into<Vec<serde_json::Value>>` or `AsRef<[serde_json::Value]>` instead so there is no need for a `vec!` at every call-site.

I think the API can be improved a little so these enqueued request/response pairs are as legible as possible.

src/transport.rs
```diff
@@ -107,24 +107,27 @@ mod tests {
...
         let response = dyn_transport.send(id, call).wait().expect("success");
         assert_eq!(response, json!(true));
 
+    #[test]
+    #[should_panic]
+    fn dyn_transport_panics() {
+        let mut transport = TestTransport::new();
+        let dyn_transport = DynTransport::new(transport.clone());
+
         // assert that the transport layer gets propagated - it errors here since
```
nlordell, line 125: nit: "it panics here"

src/transport.rs
```diff
@@ -107,24 +107,27 @@ mod tests {
...
         transport.assert_no_more_requests();
 
-        // assert that the underlying transport returns the response.
-        transport.add_response(json!(true));
         let response = dyn_transport.send(id, call).wait().expect("success");
         assert_eq!(response, json!(true));
 
+    #[test]
+    #[should_panic]
+    fn dyn_transport_panics() {
```
nlordell, line 121: I'm not sure this unit test is correct anymore, as it was supposed to test that errors were propagated. Right now it is testing `TestTransport` specific behaviour and doesn't have so much value.

Maybe `TestTransport` needs a way to queue errors as well?

nlordell (commented):
I think this is an awesome change! Just a couple of comments below as I think we can improve the `add_request_response` method a bit to make it slightly more ergonomic in tests.

fleupold (commented):
Would it make sense to use the https://github.com/asomers/mockall library here as well? This might allow us to do a convenient test transport for free.

nlordell:
> Would it make sense to use the https://github.com/asomers/mockall library here as well? This might allow us to do a convenient test transport for free.

Thats an awesome idea! Then we don't have to implement a `TestTransport` at all.

e00E:
Closing this. Will reopen later when I have more to not spam you with notifications.

PR #182, opened 2020-02-10
Fix deployed transaction signing

### LINKED ISSUE #181: [BUG] Transaction Signing is wrong for deploying contracts.
We currently RLP encode transactions for deploying contracts as if they were sending to address `0x0...0` but we should be RLP encoding it as if it had no to address (empty bytes). Take a look at the code here for how the the `to` should be RLP encoded when deploying a contract:

https://github.com/tomusdrw/rust-web3/blob/ab762e0f48c527a19952c3eb776fc7a7fb14f52e/src/api/accounts.rs#L267-L312

nlordell:
This PR fixes signing for deployed contracts. We were serializing contract deployment (a transaction with no `to`) as if it were a transaction to address 0, which is wrong.

Closes #181 

### Test Plan

New unit test generated with web3.

### DIFF
--- a/src/sign.rs
+++ b/src/sign.rs
@@ -18,7 +18,7 @@ pub struct TransactionData<'a> {
     /// Gas provided by the transaction.
     pub gas: U256,
     /// Receiver of the transaction.
-    pub to: Address,
+    pub to: Option<Address>,
     /// Value of the transaction in wei.
     pub value: U256,
     /// Call data of the transaction, can be empty for simple value transfers.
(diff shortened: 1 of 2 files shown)


fleupold (approved):
<a href="https://gitme.me/image?url=https%3A%2F%2Fmedia3.giphy.com%2Fmedia%2FtJBgiJz8bYAzuEhhjq%2Fgiphy.gif&token=zero" data-gitmeme-token="zero"></a>

PR #186, opened 2020-02-11
Add support for parsing revert errors from Partiy nodes

### LINKED ISSUE #185: Revert messages are not being correctly parsed for Parity nodes
Here is an example RPC error:
```
Rpc(Error { code: ServerError(-32015), message: "VM execution error.", data: Some(String("Reverted 0x08c379a00000000000000000000000000000000000000000000000000000000000000020000000000000000000000000000000000000000000000000000000000000002f536f6c7574696f6e7320617265206e6f206c6f6e67657220616363657074656420666f7220746869732062617463680000000000000000000000000000000000")) })
```

nlordell:
Currently revert errors from Parity nodes do not get property decoded into `ExecutionError::Revert`. Since the full nodes we run at Gnosis are Parity nodes, that is a problem. This PR adds support for that.

Fixed #185

### Test Plan

New unit tests

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -34,6 +34,7 @@ ethcontract-common = { version = "0.3.0", path = "./common" }
 ethcontract-derive = { version = "0.3.0", path = "./derive", optional = true}
 futures = { version = "0.3", features = ["compat"] }
 futures-timer = "3.0"
+hex = "0.4"
 jsonrpc-core = "14.0"
 lazy_static = "1.4"
 pin-project = "0.4"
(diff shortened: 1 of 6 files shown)


fleupold (approved):
Nice refactoring this logic into its own module. I was wondering if the geth specific decoding logic should be colocated with the error handling but since it also operates on successful results, this is probably the right way to do things.

e00E (approved):

nlordell:
> Nice refactoring this logic into its own module. I was wondering if the geth specific decoding logic should be colocated with the error handling but since it also operates on successful results, this is probably the right way to do things.

~~Good point! I think it should move as well.~~

I moved it and made the code a little harder to follow, as you would have to just to a completely new module to see how the call result gets decoded.

PR #188, opened 2020-02-12
Bump to v0.4.0

nlordell:
:point_up:

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "ethcontract"
-version = "0.3.0"
+version = "0.4.0"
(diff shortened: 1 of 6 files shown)


fleupold (approved):
Up to you to decide what makes a minor version vs a patch. Here I'd intuitively say we are just fixing some error parsing, thus maybe releasing it as 0.3.1?

nlordell:
> Up to you to decide what makes a minor version vs a patch. Here I'd intuitively say we are just fixing some error parsing, thus maybe releasing it as 0.3.1?

That would have been awesome but there were some API breaking changes, specifically with the version of web3. This would cause unexpected build errors if the old 0.3.0 version was being used.

PR #189, opened 2020-02-12
[Trivial] Add some additional commonly used re-exports in the crate root

nlordell:
:point_up:


bh2smith (approved):
Are address and H160 the same?

nlordell:
> Are address and H160 the same?

They are, in fact they are aliases of each other. However the services prefers `H160` so it would be nice for it to also be reexported

PR #203, opened 2020-02-18
`ethabi@11.0` to `ethabi@9.0` compatability layer

nlordell:
This PR decouples `ethcontract` `ethabi` version from `web3` one. This is the first step in implementing
- strategy for function name collision - this is needed as `ethabi@9.0` used by `web3` does not support multiple functions with the same name.
- ABIEncoderV2 support - `ethabi@11.0` is needed for the new `Token::Tuple` variant

Note that this will naturally go away once new `web3` versions are released but is needed for now.

### DIFF
--- a/src/errors.rs
+++ b/src/errors.rs
@@ -6,8 +6,7 @@ pub(crate) mod revert;
 mod web3contract;
 
 pub use self::web3contract::Web3ContractError;
-use ethcontract_common::abi::{Error as AbiError, ErrorKind as AbiErrorKind, Function};
-use ethcontract_common::FunctionExt;
+use ethcontract_common::abi::{Error as AbiError, Function};
 use secp256k1::Error as Secp256k1Error;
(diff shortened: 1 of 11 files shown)


e00E (approved):

src/errors.rs
```diff
@@ -39,7 +38,7 @@ pub enum DeployError {
 
     /// An error occured encoding deployment parameters with the contract ABI.
     #[error("error ABI ecoding deployment parameters: {0}")]
-    Abi(AbiErrorKind),
+    Abi(#[from] AbiError),
```
bh2smith, line 41: Never seen anything like this before.

bh2smith (approved):
Seems legit.

src/errors.rs
e00E: It's from this great crate https://github.com/dtolnay/thiserror .

PR #206, opened 2020-02-19
Fix confirmation polling forever when node skips blocks during polling

nlordell:
This PR fixes a bug with confirmation by polling (fallback for when `eth_newBlockFilter` is not supported on Infura nodes for example) where the block number could increment by more than one at a time, and because of the strict equality comparison being used, cause the transaction to never be correctly confirmed.

### Test Plan

New unit test.


fleupold (approved):
Nice, thanks for investigating this issue.

PR #220, opened 2020-03-16
Make DynTransport threadsafe

e00E:
`DynTransport` previously did not require `TransportBoxed` to be
threadsafe. This caused it to never be treated as threadsafe even when
the underlying transport was threadsafe.

### DIFF
--- a/generate/src/contract/deployment.rs
+++ b/generate/src/contract/deployment.rs
@@ -50,7 +50,7 @@ fn expand_deployed(cx: &Context) -> TokenStream {
                 F: #ethcontract::web3::futures::Future<
                     Item = #ethcontract::json::Value,
                     Error = #ethcontract::web3::Error
-                > + Send + Sync + 'static,
+                > + Send + 'static,
                 T: #ethcontract::web3::Transport<Out = F> + Send + Sync + 'static,
             {
                 use #ethcontract::contract::DeployedFuture;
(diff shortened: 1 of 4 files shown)


e00E:
I did try a `ThreadsafeTransport` but I don't think it is useful enough as you can never wrap a type that contains for example an `Rc` threadsafe (like `TestTransport`) because `Rc` is not `Send`.

generate/src/contract/deployment.rs
```diff
@@ -50,8 +50,8 @@ fn expand_deployed(cx: &Context) -> TokenStream {
                 F: #ethcontract::web3::futures::Future<
                     Item = #ethcontract::json::Value,
                     Error = #ethcontract::web3::Error
-                > + Send + 'static,
```
nlordell, line 53: nit: `Future` does not need to be `Sync`.

nlordell (approved):
LGTM

nlordell:
> I did try a ThreadsafeTransport but I don't think it is useful enough as you can never wrap a type that contains for example an Rc threadsafe (like TestTransport) because Rc is not Send.

~~It would have to wrap the transport in an `Arc<Mutex<T>>`, in which case it would become threadsafe (even an `Rc`). But I'm fine with it as is. I looked at the `TestTransport` and it doesn't really make sense to wrap the `Rc<RefCell>`s in an `Arc<Mutex>` just to promote the test transport to a thread-safe one. If, further down the line we run into issues with this, we can re-evaluate.~~

**edit**: brain-fart

nlordell:
Looks like we reached our Etherscan API limit.

generate/src/contract/deployment.rs
e00E: True. Meant to only change the Ts.

nlordell:
Configured the etherscan API key, we should be gucci now.

PR #226, opened 2020-03-20
[I256] Add bitwise operations. (4)

nlordell:
This PR implements various bit-wise and binary operations for the integer type.

### Test Plan

Most of the operations just call the `U256` implementation (`&`, `|`, etc.). For the more complex operations new unit tests were added.

### DIFF
--- a/src/int.rs
+++ b/src/int.rs
@@ -416,19 +416,29 @@ impl I256 {
 
     /// Return the least number of bits needed to represent the number.
     pub fn bits(&self) -> u32 {
-        let sign = self.signum64();
         let unsigned = self.abs_unsigned();
         let unsigned_bits = unsigned.bits();
 
-        // NOTE: We need to deal with the very special case where the number is
-        //   written as `0b11..1100..00`, in that case the number of bits
-        //   required to represent the number is equal to the number of bits
-        //   required for its two's complement. In the general case, an extra
-        //   bit is needed to represent the sign.
-        let bits = if sign == -1 && unsigned.leading_zeros() + unsigned.trailing_zeros() == 255 {
+        // NOTE: We need to deal with two special cases:
+        //   - the number is 0
+        //   - the number is a negative power of `2`. These numbers are written
+        //     as `0b11..1100..00`.
+        //   In the case of a negative power of two, the number of bits required
+        //   to represent the negative signed value is equal to the number of
+        //   bits required to represent its absolute value as an unsigned
+        //   integer. This is best illustrated by an example: the number of bits
+        //   required to represent `-128` is `8` since it is equal to `i8::MIN`
+        //   and, therefore, obviously fits in `8` bits. This is equal to the
+        //   number of bits required to represent `128` as an unsigned integer
+        //   (which fits in a `u8`).  However, the number of bits required to
+        //   represent `128` as a signed integer is `9`, as it is greater than
+        //   `i8::MAX`.  In the general case, an extra bit is needed to
+        //   represent the sign.
+        let bits = if self.count_zeros() == self.trailing_zeros() {
+            // `self` is zero or a negative power of two
             unsigned_bits
         } else {
-            unsigned_bits + (sign as usize)
+            unsigned_bits + 1
         };
 
         bits as _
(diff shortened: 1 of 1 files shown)


src/int.rs
```diff
@@ -413,6 +413,64 @@ impl I256 {
         let (result, _) = self.overflowing_neg();
         result
     }
+
+    /// Return the least number of bits needed to represent the number.
+    pub fn bits(&self) -> u32 {
```
fleupold, line 418: Should this not also return usize like U256?

src/int.rs
```diff
@@ -413,6 +413,64 @@ impl I256 {
...
+        let sign = self.signum64();
+        let unsigned = self.abs_unsigned();
+        let unsigned_bits = unsigned.bits();
+
+        // NOTE: We need to deal with the very special case where the number is
+        //   written as `0b11..1100..00`, in that case the number of bits
+        //   required to represent the number is equal to the number of bits
+        //   required for its two's complement. In the general case, an extra
+        //   bit is needed to represent the sign.
+        let bits = if sign == -1 && unsigned.leading_zeros() + unsigned.trailing_zeros() == 255 {
```
fleupold: I don't understand this case. Shouldn't there just be a single 1 other than the sign bit? In the example I see 4 ones. Also, why can we omit the sign bit in this case?

src/int.rs
```diff
@@ -803,4 +948,24 @@ mod tests {
         assert_eq!(-(-I256::MAX), I256::MAX);
         assert_eq!(I256::MIN.checked_neg(), None);
     }
+
+    #[test]
+    fn bits() {
```
fleupold, line 975: Are you testing the special case somewhere here?

src/int.rs
```diff
@@ -558,6 +616,93 @@ impl ops::Not for I256 {
...
+        impl ops::ShrAssign<$t> for I256 {
+            fn shr_assign(&mut self, rhs: $t) {
+                *self = *self >> rhs;
+            }
+        }
+    };
+}
+
+impl_shift!(i8, u8, i16, u16, i32, u32, i64, u64, i128, u128, isize, usize);
+impl_shift!(I256, U256[from_raw]);
```
fleupold, line 726: what does `U256[from_raw]` mean?

fleupold (approved):
Seems fine. I don't understand the bits special case.

src/int.rs
nlordell: I changed it to be more inline with standard rust types. For example, another method that operates on bits: https://doc.rust-lang.org/std/primitive.i32.html#method.count_zeros

src/int.rs
nlordell: Yes, `i64::MIN` as well as `-0b1000`.

nlordell: Also, without the special case `I256::MIN.bits() == 257`.

src/int.rs
nlordell: Its because `I256: !From<U256>` because of potential overflows. I handle the special syntax in the macro here

https://github.com/gnosis/ethcontract-rs/blob/i256-4/src/int.rs#L667-L670

By default, types will use `from` conversion, but `U256` needs to use `from_raw` conversion.

src/int.rs
nlordell: I guess I didn't explain it too well :cry:. Its easier seen in an example: without this special case:
- `I256::from(i8::MIN).bits() == 9`
- `I256::from(i64::MIN).bits() == 65`
- ...
- `I256::MIN.bits() == 257`

It comes from the fact that two's complement representation doesn't actually use a full bit to encode a sign, and uses the fact that 0 has no sign to get 1 extra value.

src/int.rs
fleupold: 🙆‍♂

src/int.rs
fleupold: Thanks, I get it now, but I still think the example in the comment is wrong:

`0b11..1100..00` -> shouldn't this either be `0b0...1....0` or (if before the complement) `0b1....10` (not sure if BE or LE)

nlordell: `i8::MIN` as a `I256` is represented as `0b11..1110000000`, and its in those specific cases that we want to not add 1 to the required number of bits. The absolute value of `i8::MIN` is `0b00..0010000000`. So we don't add 1 if it is a negative number and the number of zeros == the number of trailing zeros, or the sum of trailing zeros and leading zeros of the absolute value is 255 (i.e. the absolute value is a power of 2).

Maybe a better way to explain it is that the number of bits does not need the +1 if it is a negative power of two (because of how two's complement works).

nlordell: @fleupold I updated the comment and the condition. Let me know if it makes more sense now.

fleupold: Yes, it's very clear to me now. Thanks for updating the explanation.

e00E (approved):

PR #229, opened 2020-03-20
[I256] Implement JSON RPC traits. (7)

nlordell:
Implement some utility traits that are required for JSON RPC. Namely, serialization and tokenization.

### Test Plan

Unit Tests

PR #224, opened 2020-03-20
[I256] Adds conversions to and from standard integer types. (2)

nlordell:
Adds to and from conversions from standard integer types and `U256`.

### Test Plan

Unit tests

PR #231, opened 2020-03-21
[I256] Integrate new 256-bit integer type with generated code. (9)

nlordell:
At last! This PR implements a 256-bit signed integer type and integrates it with `ethcontract` for contracts that return signed integer types.

### Info

This PR will be merged directly into `master`. This closes #223, #224, #225, #226, #227, #228, #229, #230 which were created to make this large feature easier to review.

### Test Plan

Unit tests and an new example that gets executed with CI, run with:
```sh
(cd examples/truffle; yarn && yarn build && yarn start) # separate terminal
cargo run --example abi
```

Should get output like:
```
   Compiling ethcontract v0.5.0 (/var/home/nlordell/Developer/ethcontract-rs)
    Finished dev [unoptimized + debuginfo] target(s) in 3.89s
     Running `target/debug/examples/abi`
Using contract at 0xcd38694071e9f2b669ce896d17ca54ad327e3605
get_u8() -> u8
  ⏎ 231
get_u16() -> u16
  ⏎ 25063
get_u32() -> u32
  ⏎ 3593167335
get_u64() -> u64
  ⏎ 17214110251719811559
get_u128() -> u128
  ⏎ 226019319247017660072433873994988085735
get_u256() -> primitive_types::U256
  ⏎ 56251707932527615379356343136198624385371605568657930206938880970442100400615
get_i8() -> i8
  ⏎ -25
get_i16() -> i16
  ⏎ 25063
get_i32() -> i32
  ⏎ -701799961
get_i64() -> i64
  ⏎ -1232633821989740057
get_i128() -> i128
  ⏎ -114263047673920803390940733436780125721
get_i256() -> ethcontract::int::I256
  ⏎ 56251707932527615379356343136198624385371605568657930206938880970442100400615
get_bool() -> bool
  ⏎ true
get_bytes() -> alloc::vec::Vec<u8>
  ⏎ [214, 43, 97, 231]
get_fixed_bytes() -> [u8; 6]
  ⏎ [206, 17, 214, 43, 97, 231]
get_address() -> primitive_types::H160
  ⏎ 0x782b08f9aa09bccc754e62c5eee4ce11d62b61e7
get_string() -> alloc::string::String
  ⏎ "eee4ce11d62b61e7"
get_array() -> alloc::vec::Vec<u64>
  ⏎ [8961414089469494464, 13648014267371948281, 12252531847565566661, 17214110251719811559]
get_fixed_array() -> [i32; 3]
  ⏎ [1968071365, -286994927, -701799961]
```


fleupold (approved):
👏 

⛵️

e00E (approved):

src/int.rs
```diff
@@ -0,0 +1,1409 @@
...
+
+    /// The maximum value which can be inhabited by this type.
+    pub const fn max_value() -> Self {
+        I256::MAX
+    }
+
+    /// The minimum value which can be inhabited by this type.
+    pub const fn min_value() -> Self {
+        I256::MIN
+    }
```
bh2smith, line 96: Why introduce these if we already have the public constants?

src/int.rs
```diff
@@ -0,0 +1,1409 @@
...
+use std::str;
+use std::{i128, i64, u64};
+use web3::contract::{self, tokens};
+use web3::types::U256;
+
+/// Compute the two's complement of a U256.
+fn twos_complement(u: U256) -> U256 {
+    let (twos_complement, _) = (!u).overflowing_add(U256::one());
+    twos_complement
+}
```
bh2smith, line 19: This is so interesting! Today I learned,

https://en.wikipedia.org/wiki/Two%27s_complement

After studying pure mathematics for decades, there are still so many cute little tricks used in computer science that I have never seen before. This file alone contains several of them.

src/int.rs
```diff
@@ -0,0 +1,1409 @@
...
+            match self.pow_sign(exp) {
+                Sign::Positive => I256::MAX,
+                Sign::Negative => I256::MIN,
+            }
+        } else {
+            result
+        }
+    }
+
+    /// Wrapping powolute value. Computes `self.pow()`, wrapping around at the
```
bh2smith, line 723: What is "powolute"? Is this some combination of power and absolute? Absolute Power. If I remember correctly, this was Jafar's last wish that turned him into a genie and saved the day for everyone else.

src/int.rs
```diff
@@ -0,0 +1,1409 @@
...
+    ///
+    /// Returns a tuple of the exponentiation along with a bool indicating
+    /// whether an overflow happened.
+    pub fn overflowing_pow(self, exp: u32) -> (Self, bool) {
+        let sign = self.pow_sign(exp);
+        let (unsigned, overflow_pow) = self.abs_unsigned().overflowing_pow(exp.into());
+        let (result, overflow_conv) = I256::overflowing_from_sign_and_abs(sign, unsigned);
+
+        (result, overflow_pow || overflow_conv)
+    }
```
bh2smith, line 697: I need to understand the overflowing better... At first I assumed the overflowing power would be the one that wraps around, but then I see you have a `wrapping_pow` below. This can only mean that this overflowing power actually gives the correct result in the event that the numbers we unbounded. However, this would mean that the result could fall out of scope and I would no longer know what type to expect... 😕 ❓

src/int.rs
```diff
@@ -0,0 +1,1409 @@
...
+            result
+        }
+    }
+
+    /// Wrapping powolute value. Computes `self.pow()`, wrapping around at the
+    /// boundary of the type.
+    pub fn wrapping_pow(self, exp: u32) -> Self {
+        let (result, _) = self.overflowing_pow(exp);
+        result
+    }
```
bh2smith, line 728: I guess this would be the place where one could implement `{overflowing, checked, saturating, wrapping}` `div` and `mod`. At first glance though, I can't see how division could ever overflow... since division only "blows up" when dividing by real numbers strictly between 0 and 1.

bh2smith (commented):
This is awesome. I have included a few questions and comments in `int.rs`

src/int.rs
nlordell: Just for consistency with primitive integer types where both the [constant exists](https://doc.rust-lang.org/std/u32/constant.MAX.html) as well as a [const method](https://doc.rust-lang.org/std/primitive.u32.html#method.max_value).

src/int.rs
nlordell: :man_facepalming: That is yet another typo.

src/int.rs
nlordell: - `overflowing_*` methods compute the result of the operation and return the value and returns both the value, that is wrapped in case of an overflow, and a bool indicating whether or not an overflow occured.
- `wrapping_*` return the exact same result value, but just ignore the overflow bool.
- `checked_*` retruns `Some(value)` when the result fits in the signed integer (i.e. no overflow) and `None` if there is an overflow.
- `saturating_*` returns the value result of the operation if the result fits in the signed integer and `I256::MAX` or `I256::MIN` depending on the direction of the overflow. 

Hope this explains it a little better.

src/int.rs
nlordell: `I256::MIN / -1` would overflow, as `-I256::MIN` also overflows. The other subtlety is that `checked_div` returns a `None` when dividing by 0, while the other operations should panic. (This is just to maintain consistency with the signed primitive integers such as `i32`.

src/int.rs
nlordell: I had done a search and replace all of `abs` for `pow` and this comment got caught in the crossfire :stuck_out_tongue_closed_eyes:

src/int.rs
bh2smith: Awesome, so it does make sense to implement all four versions for each then for consistency even if they are only single cases for some.

nlordell: When in doubt, I just look at what the standard library does with its signed integers: https://doc.rust-lang.org/std/primitive.i32.html

It includes nice documentation and examples that illustrate some of these edge cases.

nlordell: Usually (take a look at the other operations `+ - *`) all of the variants can be implemented in terms of the `overflowing_*` variant of the operation.

bh2smith: Yes! thanks for the sick hints.

PR #233, opened 2020-03-24
Decoding for typed events

nlordell:
This PR adds decoding to the `ethcontract::contract::Instance` type for events. That is, logs for the matching events get decoded into the parameter types.

### Test Plan

New unit test and example code now uses the typed events.


examples/events.rs
```diff
@@ -29,8 +30,9 @@ async fn run() {
         .expect("deployment failed");
     let mut transfers = instance
         .instance
-        .event("Transfer")
-        .expect("transfer event not found");
+        .event::<_, (Address, Address, U256)>("Transfer")
```
e00E, line 33: I imagined that we would parse the event type from the contract like `event Deposit(address indexed user, address indexed token, uint256 amount, uint32 batchId);` and then I would get this as a rust type. Then I could say *give me all events* and get an `enum { Deposit(...), Withdraw(...), ... }`.

Is that something you had in mind for the future?

e00E (approved):

examples/events.rs
nlordell: That would require code generation. This is just the runtime implementation for decoding topics + logs for a given event.

Also, that would be the equivalent of the `all_events` and not `event("Name")`.

nlordell: To answer your question though, yes that is planned.

src/contract.rs
```diff
@@ -144,28 +146,28 @@ impl<T: Transport> Instance<T> {
...
-            filter,
-            std::time::Duration::from_secs(1),
+            event.clone(),
+            self.address(),
         ))
     }
+
+    /// Returns a log stream that emits a log for every new event emitted after
+    /// the stream was created for this contract instance.
+    pub fn all_event(&self) -> LogStream<T> {
```
fleupold: nit: all_event**s** ?

src/contract/event.rs
```diff
@@ -0,0 +1,298 @@
...
+            Event::Removed(value) => value,
+        }
+    }
+
+    /// Get the underlying event data if the event was added.
+    ///
+    /// # Panics
+    ///
+    /// This method panics if the instance is a removed event.
+    pub fn added(self) -> T {
```
fleupold: Curious to hear your thoughts why panicking is better than making the signature return Option<T>?

src/contract/event.rs
```diff
@@ -0,0 +1,298 @@
...
+    /// This method panics if the instance is a added event.
+    pub fn removed(self) -> T {
+        match self {
+            Event::Removed(value) => value,
+            Event::Added(_) => panic!("attempted to unwrap an added event to a removed one"),
+        }
+    }
+}
+
+/// The default poll interval to use for confirming transactions.
```
fleupold: for confirming transactions?

src/contract/event.rs
```diff
@@ -0,0 +1,298 @@
...
+#[cfg(not(test))]
+pub const DEFAULT_POLL_INTERVAL: Duration = Duration::from_secs(5);
+#[cfg(test)]
+pub const DEFAULT_POLL_INTERVAL: Duration = Duration::from_secs(0);
+
+/// A builder for creating a filtered stream of contract events that are
+#[must_use = "event builder do nothing unless you or stream them"]
```
fleupold: you **use** or stream them?

fleupold (approved):
🌲

src/contract/event.rs
nlordell: Oops!

src/contract/event.rs
nlordell: Thanks!

src/contract/event.rs
nlordell: :+1:

src/log.rs
```diff
@@ -19,6 +19,7 @@ use web3::Transport;
 /// Note that when creating a log stream that is only valid until a certain
 /// block number, the `Stream` implementation will currently remain in the
 /// pending state indefinitely.
+#[must_use = "streams do nothing unless you or poll them"]
```
bh2smith, line 22: "... unless you ??? or poll them"

PR #239, opened 2020-03-26
[Easy] Improve error messages on compile errors

nlordell:
Use debug formatting for error messages on compile errors. This provides better error messages that look like this now:

```
error: top level error

Caused by:
    0: error bubbled up once
    1: Root error
```

This PR also removes the JSON blob from the error message as it made the error unreadable.


e00E (approved):

generate/src/contract.rs
```diff
@@ -138,7 +138,7 @@ impl Default for Context {
 
 pub(crate) fn expand(args: Args) -> Result<TokenStream> {
     let cx = Context::from_args(args)?;
-    let contract = expand_contract(&cx).context("error expanding contract from source")?;
+    let contract = expand_contract(&cx).context("error expanding contract from its ABI")?;
```
bh2smith, line 141: This change appears to only be replacing the word "source" with "its ABI" in an an error string and does not appear to have anything to do with the PR's description.

nlordell: Hmm.. It appears to be because I merged things in the wrong order. This was the original change:
https://github.com/gnosis/ethcontract-rs/pull/239/commits/251066badd8c55dffbba69111a54d4076f1cdbf4

PR #242, opened 2020-03-27
Use succinct type names for generated code

nlordell:
This PR changes the names of types in generated code from `#contract_name` to `Contract` and `#{contract_name}Methods` to `Methods`. This is done for a few reasons:
1. Code generation is simpler as there are less variables - the type names are always the same so there are fewer worries about name collisions
2. The simplified names are nicer to read and contain less redundant naming:
   - `my_cool_contract::MyCoolContractMethods` is now `my_cool_contract::Methods`
   - `my_cool_contract::MyCoolContract` is now `my_cool_contract::Contract`, which is aliased to `MyCoolContract`.
3. We can now import the runtime crate as a known name in the root of the module, simplifying code that previously needed the runtime crate name as a variable.
4. All types in the root of the generated module have fixed names, making duplicate name issues less likely and easier to fix in the future.

Note that the re-exported contract type is still called `#contract_name` so current code does not have to be changed.

### Test Plan

Examples still compile without change


e00E (approved):

fleupold (approved):
So nice!

🌈

bh2smith (commented):
Fabulous 🎈

PR #246, opened 2020-03-27
Use 32-bit method selector instead of function signature

nlordell:
Use 32-bit method selector instead of function signature for method identification in the runtime.

### Test Plan

CI

### DIFF
--- a/common/src/hash.rs
+++ b/common/src/hash.rs
@@ -14,7 +14,10 @@ where
     output
 }
 
-/// A 32-byte hash prefix.
+/// A 32-bit prefix of a standard 256-bit Keccak hash.
+///
+/// This 32-bit prefix is generally used as the first 4 bytes of transaction
+/// data in order to select which Solidity method will be called.
 pub type H32 = [u8; 4];
 
 /// Calculate the function selector as per the contract ABI specification. This
--- a/common/src/hash.rs
+++ b/common/src/hash.rs
@@ -14,15 +14,18 @@ where
     output
 }
 
+/// A 32-byte hash prefix.
+pub type H32 = [u8; 4];
+
 /// Calculate the function selector as per the contract ABI specification. This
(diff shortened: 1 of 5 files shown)


e00E (approved):
I don't see anything wrong but I also don't understand the reason for this change.

nlordell:
@e00E Its for consistency with events. Additionally, when you debug the `Instance` you will see all the event selectors now which might be able to help with debugging web3 calls.

fleupold (approved):
Isn't it less convenient for debugging to not see the full function signature but only the first bytes of the hash? Or are we still using the clear text name for error propagation?

E.g:

> ERRO [driver::driver::scheduler::system] Batch 5284224 failed with unretryable error: method 'submitSolution(uint32,uint256,address[],uint16[],uint128[],uint128[],uint16[]):(uint256)' failure: contract call reverted with message: Some("Solutions are no longer accepted for this batch")

nlordell:
Errors come from the full method signature. This is just used for the method lookup table. The reason we need to method lookup table is that the `ethabi::Contract` API assumes that there are no duplicate methods and just returns the first method whose name matches the search parameters.

I had originally done this with method signature, but I thought, for consistency with how we do it for events, we should use the hash selector.

Also, this has the added benefit of `println!("{:?}", instance)` dumping all the method selectors.

common/src/hash.rs
```diff
@@ -14,15 +14,18 @@ where
     output
 }
 
+/// A 32-byte hash prefix.
```
bh2smith: The word prefix here suggests to me that this is only the beginning of a 32-byte hash (like the `0x` to indicate it is a hex representation). Is this intended?

nlordell: Yes, it is the first 4 bytes in a 32-byte keccak hash. I'll clarify in the comments.

PR #248, opened 2020-03-27
[For Fun] Added support for HTTP and npmjs contracts

nlordell:
Added support for additional contract sources.

### Test Plan

Example that uses npmjs package source.

### DIFF
--- a/derive/src/lib.rs
+++ b/derive/src/lib.rs
@@ -31,13 +31,17 @@ use syn::{
 /// ethcontract::contract!("build/contracts/MyContract.json");
 /// ```
 ///
-/// Alternatively, an etherscan URL can be specified. In this case the ABI will
-/// be retrieved and used to generate type-safe contract bindings:
+/// Alternatively, an other sources may be used, for full defailts consult the
+/// `ethcontract-generate::source` documentation. Some basic examples:
 ///
 /// ```ignore
+/// // HTTP(S) source
+/// ethcontract::contract!("https://my.domain.local/path/to/contract.json")
+/// // Etherscan.io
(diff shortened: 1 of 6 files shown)


fleupold (approved):
Super cool. Should we change dex-contracts to use that instedad of checking out the git submodule?

e00E (approved):

derive/src/lib.rs
```diff
@@ -31,13 +31,17 @@ use syn::{
 /// ethcontract::contract!("build/contracts/MyContract.json");
 /// ```
 ///
-/// Alternatively, an etherscan URL can be specified. In this case the ABI will
-/// be retrieved and used to generate type-safe contract bindings:
+/// Alternatively, an other sources may be used, for full defailts consult the
```
bh2smith, line 34: 1. "an other sources" should be "an other source" or "other sources". Since we want plural here I would think removing the "an" should work.

2. typo on "defailts"

derive/src/lib.rs
```diff
@@ -31,13 +31,17 @@ use syn::{
...
 ///
 /// ```ignore
+/// // HTTP(S) source
+/// ethcontract::contract!("https://my.domain.local/path/to/contract.json")
+/// // Etherscan.io
 /// ethcontract::contract!("etherscan:0x0001020304050607080910111213141516171819");
-/// // or
 /// ethcontract::contract!("https://etherscan.io/address/0x0001020304050607080910111213141516171819");
+/// // npmjs
+/// ethcontract::contract!("npm:@org/package@1.0.0/path/to/contract.json")
```
bh2smith, line 44: Very cool!

PR #257, opened 2020-03-30
Attach transaction information to contract instance

nlordell:
This PR adds a `transaction_hash` field to contract instances that keeps track (when it is available) of the transaction used to deploy the contract. This is useful for querying the block number used to deploy this transaction so that when streaming events, the starting block can be more easily and programmatically determined.

### Test Plan

New assertion in unit test and Rinkeby example now logs transaction hash.


e00E (approved):

fleupold (approved):
Looks good, although performance wise I'm not sure if you specify fromBlock:0 or fromBlock:deployment when querying events.

nlordell:
> Looks good, although performance wise I'm not sure if you specify fromBlock:0 or fromBlock:deployment when querying events.

That's a good point... me neither. I still think its nice extra information about a contract to have that comes mostly for free.

PR #259, opened 2020-03-31
Added some very basic fallback function support

nlordell:
Added some basic fallback function support for completeness. This was part of milestone 0.6.0 which we about to release (with the event support) and I think this minimal implementation is OK for now.

### Test Plan

Example uses fallback function now!


fleupold (approved):
🍂🔙

e00E (approved):

PR #260, opened 2020-03-31
Implemented rustfmt support for the generated code

nlordell:
:point_up: 

Works better with things like `racer` as well as makes debugging issues with generated code a bit easier!

### Test Plan

Enabled by default and tested by an example.

### DIFF
--- a/generate/src/contract.rs
+++ b/generate/src/contract.rs
@@ -154,6 +154,7 @@ fn expand_contract(cx: &Context) -> Result<TokenStream> {
     Ok(quote! {
         #[allow(dead_code)]
         #vis mod #contract_mod {
+            #[rustfmt::skip]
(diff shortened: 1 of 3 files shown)


fleupold (approved):
💾

e00E (approved):
I wonder if it is possible to use rustfmt as a library (in the future) to avoid shelling out to a binary.

nlordell:
> I wonder if it is possible to use rustfmt as a library (in the future) to avoid shelling out to a binary.

There is `rustfmt_lib` but that would require compiling a lot of extra stuff. I think, as a future improvement, have it be a optional feature to compile `rustfmt` into the `ethcontract-generate` instead of shelling out.

I'd also be concerned about increased compile times - but its worth investigating :smile:

PR #263, opened 2020-04-01
Include past logs in the stream

e00E:
No need to review yet.

Calling `stream` on the filter translates to a `eth_getFilterChanges`
rpc call which does not include past logs from before the filter's
creation.

This is just the first version I got working to be able to test the event based orderbook. I am sharing it because @nlordell and I talked about it.

### DIFF
--- a/src/contract/event.rs
+++ b/src/contract/event.rs
@@ -55,7 +55,7 @@ pub struct EventMetadata {
     /// The index of the log in the block.
     pub log_index: usize,
     /// The log index in the transaction this log belongs to.
-    pub transaction_log_index: usize,
+    pub transaction_log_index: Option<usize>, // TODO: this wasn't available in my test
     /// The log type. Note that this property is non-standard but is supported
     /// by Parity nodes.
     pub log_type: Option<String>,
@@ -121,7 +121,7 @@ impl EventMetadata {
             transaction_hash: log.transaction_hash?,
             transaction_index: log.transaction_index?.as_usize(),
             log_index: log.log_index?.as_usize(),
-            transaction_log_index: log.transaction_log_index?.as_usize(),
+            transaction_log_index: log.transaction_log_index.map(|i| i.as_usize()),
             log_type: log.log_type.clone(),
         })
     }
--- a/src/log.rs
+++ b/src/log.rs
@@ -2,6 +2,7 @@
 //! emitted by a contract.
 
 use crate::errors::ExecutionError;
+use crate::future::CompatCallFuture;
 use futures::compat::{Compat01As03, Future01CompatExt, Stream01CompatExt};
 use futures::ready;
 use futures::stream::Stream;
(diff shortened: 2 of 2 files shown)


src/contract/event.rs
```diff
@@ -55,7 +55,7 @@ pub struct EventMetadata {
     /// The index of the log in the block.
     pub log_index: usize,
     /// The log index in the transaction this log belongs to.
-    pub transaction_log_index: usize,
+    pub transaction_log_index: Option<usize>, // TODO: this wasn't available in my test
```
nlordell, line 58: Hmm, odd, we should probably dig a little and see why.

src/log.rs
```diff
@@ -55,12 +67,50 @@ impl<T: Transport> Stream for LogStream<T> {
...
+                LogStreamState::StreamingPastLogs {
+                    poll_interval,
+                    filter,
+                    past_logs,
+                } => {
+                    if let Some(log) = past_logs.pop() {
+                        return Poll::Ready(Some(Ok(log)));
+                    } else {
+                        // TODO: this could duplicate some logs if they appear in both getFilterLogs
+                        // and getFilterChanges or it could skip some events.
```
nlordell, line 110: I would hope that `eth_getFilterLogs` would count as the "last polled" call when getting the `eth_getFilterChanges` :crossed_fingers:

src/log.rs
```diff
@@ -112,17 +162,26 @@ mod tests {
...
-            .expect("log stream did not produce any logs")
-            .expect("failed to get log from log stream");
-
-        assert_eq!(log.log_type.as_deref(), Some("awesome"));
+        transport.add_response(json!([generate_log("awesome1")]));
+
+        let mut stream = LogStream::new(web3, Default::default(), Duration::from_secs(0));
+        let mut logs = Vec::new();
+        for _ in 0..2 {
+            let log = stream
```
nlordell, line 173: You might be able to use some stream combinators to make this feel more like iterators.

src/log.rs
```diff
@@ -55,12 +67,50 @@ impl<T: Transport> Stream for LogStream<T> {
...
+                        Ok(filter) => filter,
                         Err(err) => return Poll::Ready(Some(Err(err.into()))),
                     };
-                    let stream = log_filter.stream(*poll_interval).compat();
-                    LogStreamState::Streaming(stream)
+                    // TODO: this request can be very slow and the response very big. We might want
+                    // to be smarter about it. However, there is no way to page this request. Maybe
+                    // Set up intermediate queries to handle a limited number of blocks at a time.
+                    let past_logs_future = CompatCallFuture::<T, Vec<Log>>::new(filter.logs());
+                    LogStreamState::GettingPastLogs {
```
nlordell, line 78: I think that we shouldn't enter this state unconditionally. I think that if `from_block` is `latest` or `pending` we can skip straight to streaming.

nlordell (commented):
This is starting to look pretty good to me.

PR #264, opened 2020-04-02
Returned boxed transaction receipt on failure

nlordell:
:point_up: This allows callers to inspect the transaction receipt when it failed because of a revert or out of gas (for example). 

Note that the transaction receipt is boxed - this is because it it a rather sizeable struct and it doesn't make sense for the error types, and therefore the result types, to be so large. Not boxing also produces a clippy lint warning.

### Test Plan

Nothing to do! This error variant was always created from a transaction receipt anyway so there is no new logic.

### DIFF
--- a/src/contract/deploy.rs
+++ b/src/contract/deploy.rs
@@ -200,9 +200,7 @@ where
         let address = match tx.contract_address {
             Some(address) => address,
             None => {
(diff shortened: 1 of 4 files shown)


fleupold (approved):
Swift & sweet!

bh2smith (commented):
schwifty! 🏊

PR #266, opened 2020-04-02
Allow passing additional derives to event structs and enums

e00E:
This is useful for example to make them derive serde serialization.

### DIFF
--- a/derive/src/lib.rs
+++ b/derive/src/lib.rs
@@ -19,8 +19,8 @@ use std::error::Error;
 use syn::ext::IdentExt;
 use syn::parse::{Error as ParseError, Parse, ParseStream, Result as ParseResult};
 use syn::{
-    braced, parenthesized, parse_macro_input, Error as SynError, Ident, LitInt, LitStr, Token,
-    TypePath, Visibility,
+    braced, parenthesized, parse_macro_input, Error as SynError, Ident, LitInt, LitStr, Path,
+    Token, Visibility,
 };
 
 /// Proc macro to generate type-safe bindings to a contract. This macro accepts
@@ -139,11 +139,9 @@ impl ContractArgs {
                 Parameter::Methods(methods) => methods.into_iter().fold(builder, |builder, m| {
                     builder.add_method_alias(m.signature, m.alias)
                 }),
-                Parameter::EventDerives(derives) => {
-                    derives.into_iter().fold(builder, |builder, derive| {
-                        builder.add_event_derive(derive.to_token_stream().to_string())
-                    })
-                }
+                Parameter::EventDerives(derives) => derives
+                    .into_iter()
+                    .fold(builder, |builder, derive| builder.add_event_derive(derive)),
             };
         }
 
@@ -195,7 +193,7 @@ enum Parameter {
     Crate(String),
     Deployments(Vec<Deployment>),
     Methods(Vec<Method>),
-    EventDerives(Vec<TypePath>),
+    EventDerives(Vec<String>),
 }
 
 impl Parse for Parameter {
@@ -276,9 +274,10 @@ impl Parse for Parameter {
                 let content;
                 parenthesized!(content in input);
                 let derives = content
-                    .parse_terminated::<_, Token![,]>(TypePath::parse)?
+                    .parse_terminated::<_, Token![,]>(Path::parse)?
                     .into_iter()
-                    .collect::<Vec<_>>();
+                    .map(|path| path.to_token_stream().to_string())
+                    .collect();
                 Parameter::EventDerives(derives)
             }
             _ => {
@@ -403,10 +402,6 @@ mod tests {
         }
     }
 
-    fn derive(derive: &str) -> TypePath {
-        syn::parse_str::<TypePath>(derive).unwrap()
-    }
-
     #[test]
     fn parse_contract_args() {
         let args = contract_args!("path/to/artifact.json");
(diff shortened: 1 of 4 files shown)


fleupold (approved):
Looks good to me! Passing in arbitrary derive logic seems a bit intrusive, so I'm wondering if this kind of thing is a standard feature for other code generation crates?

How does it work if you derive traits from libraries that ethcontracts doesn't itself depend on? Also, are all events serializable, deserializable? If so, should we maybe natively derive them.

derive/src/lib.rs
```diff
@@ -187,6 +195,7 @@ enum Parameter {
     Crate(String),
     Deployments(Vec<Deployment>),
     Methods(Vec<Method>),
+    EventDerives(Vec<TypePath>),
```
nlordell: ubernit: I would move the `to_string()` to the parsing of the parameter types, just to keep it consistent with the other parameters, where the data is 1:1 what the builder is expecting.

derive/src/lib.rs
```diff
@@ -263,6 +272,15 @@ impl Parse for Parameter {
 
                 Parameter::Methods(methods)
             }
+            "event_derives" => {
+                let content;
+                parenthesized!(content in input);
+                let derives = content
+                    .parse_terminated::<_, Token![,]>(TypePath::parse)?
```
nlordell: This should be a `Path` and not a `TypePath`, type paths allow for `<Foo as Bar>::Type` which is not allowed for `derive`s AFAIK.

generate/src/contract.rs
```diff
@@ -18,7 +18,7 @@ use inflector::Inflector;
 use proc_macro2::{Ident, Literal, TokenStream};
 use quote::quote;
 use std::collections::HashMap;
-use syn::Visibility;
+use syn::{TypePath, Visibility};
```
nlordell: Should be a `Path` and not a `TypePath`.

e00E:
> Looks good to me! Passing in arbitrary derive logic seems a bit intrusive, so I'm wondering if this kind of thing is a standard feature for other code generation crates?

I don't know how other crates do it but something like this is the only way to give the user of the library the most freedom. For macros that create structs you can usually use your own derives just by putting them in front of the struct. In ethcontract it's more complicated because we generate several structs.

> How does it work if you derive traits from libraries that ethcontracts doesn't itself depend on?

Shouldn't matter because ethcontract just generates the code for the contract, but does not compile. 
The code that then uses the generated code has to have the necessary  dependencies like serde.

> Also, are all events serializable, deserializable?

I am not sure but I believe all underlying types are serializable.

> If so, should we maybe natively derive them.

If we unconditionally derive them then it means that every generated contract depends on serde. It gives the users of the library more freedom to decide on their own whether they want the dependency and extra compile time.

generate/src/contract.rs
```diff
@@ -104,6 +106,13 @@ impl Context {
             }
         }
 
+        let event_derives = args
+            .event_derives
+            .iter()
+            .map(|derive| syn::parse_str::<TypePath>(derive))
+            .collect::<Result<Vec<TypePath>, syn::Error>>()
```
nlordell: ubernit: the types can be determined by the compiler:
```suggestion
            .collect::<Result<Vec<_>, _>>()
```
I don't think being explicit here adds value especially since we coerse all errors into `anyhow::Error`s and the `Vec` item type is on the line right above.

generate/src/contract/events.rs
```diff
@@ -5,7 +5,8 @@ use ethcontract_common::abi::{Event, EventParam, Hash};
 use ethcontract_common::abiext::EventExt;
 use inflector::Inflector;
 use proc_macro2::{Literal, TokenStream};
-use quote::quote;
+use quote::{quote, ToTokens as _};
```
nlordell: I don't think `ToTokens` trait is needed.

generate/src/contract/events.rs
```diff
@@ -42,10 +43,18 @@ fn expand_structs_mod(cx: &Context) -> Result<TokenStream> {
     })
 }
 
+fn expand_derives(derives: &[TypePath]) -> TokenStream {
+    let derives = derives
+        .iter()
+        .map(|derive| derive.to_token_stream())
+        .collect::<Vec<_>>();
+    quote! {#(#derives),*}
```
nlordell: nit: `ToTokens::to_token_stream()` gets called automatically for you by `quote!`.

```suggestion
    quote! {#(#derives),*}
```

nlordell (approved):
Awesome! Just one small change, as derive types should be `Path` and not `TypePath`, and a few really small nits.

nlordell:
> Looks good to me! Passing in arbitrary derive logic seems a bit intrusive, so I'm wondering if this kind of thing is a standard feature for other code generation crates?

When deriving a single type, I think it's pretty common (at least I've seen it in other code generating macros/proc_macros) to allow arbitrary attributes, including `#[derive(...)]`. I think this is pretty close to standard and is nice given the nature of the generated types.

> Shouldn't matter because ethcontract just generates the code for the contract, but does not compile. The code that then uses the generated code has to have the necessary dependencies like serde.

:+1: 

>  Also, are all events serializable, deserializable?

`Serialize` and `Deserialize` can be trivially implemented for them all

> If so, should we maybe natively derive them.

I'd be OK with that, but I think that this feature is super valuable regardless if we default to implementing `Serialize` and `Deserialize` as it allows for any auto-trait to be derived.

> If we unconditionally derive them then it means that every generated contract depends on serde. It gives the users of the library more freedom to decide on their own whether they want the dependency and extra compile time.

Serde is already a dependency of `ethcontract`, so we could just re-export it `in ethcontract::private` and use it from there, that way no new dependencies would be required. That being said, I tend to agree with @e00E here, and I think that implementing `Serialize` and `Deserialize` is kind of specific and not a general thing that most library users would need.

PR #270, opened 2020-04-03
Add limit param to event builders

nlordell:
Adds the non-standard `limit` filter parameter to the event builders.

### Test Plan

Unit test.

### DIFF
--- a/src/contract/event.rs
+++ b/src/contract/event.rs
@@ -205,6 +205,14 @@ impl<T: Transport, E: Detokenize> EventBuilder<T, E> {
         self
     }
 
+    /// Limit the number of events that can be retrieved by this filter.
+    ///
+    /// Note that this parameter is non-standard.
+    pub fn limit(mut self, value: usize) -> Self {
+        self.filter = self.filter.limit(value);
+        self
+    }
+
     /// Adds a filter for the first indexed topic.
     ///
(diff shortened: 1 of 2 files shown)


src/contract/event.rs
```diff
@@ -465,6 +473,14 @@ impl<T: Transport, E: ParseLog> AllEventsBuilder<T, E> {
...
     }
 
+    /// Limit the number of events that can be retrieved by this filter.
+    ///
+    /// Note that this parameter is non-standard.
+    pub fn limit(mut self, value: usize) -> Self {
+        self.filter = self.filter.limit(value);
+        self
+    }
+
```
bh2smith, line 483: I feel like this entire block is duplicated...

bh2smith (approved):
Looks good, but I see an entire block of what appears to be duplicated code.

src/contract/event.rs
nlordell: It's unfortunate but we currently need two separate types for building a filter for all events and for a single event, since they have slightly different semantics.

Here is a link to the original discussion: https://github.com/gnosis/ethcontract-rs/pull/250#discussion_r400055786

While I plan on improving on this code in the future, it is a necessary evil for now :cry:. A code smell if you will.

e00E (approved):

fleupold (approved):

Issue #290, opened 2020-04-27
Log And Event Stream Refactor


nlordell:
Refactoring of the log and event stream in `ethcontract` for:
- [x] `eth_getLogs` based event stream that tracks re-orgs client side. This allows the event stream to work behind load balancers and be fault-tolerant in that it continues working even if a node were to go down during event streaming
- [x] Join `EventBuilder` and `AllEventBuilder` types, as there is a fair amount of duplicated logic that would be nice to remove.
- [x] Change the paginated query to be a `Stream` of pages instead of `Future` where they are all collected - this would allow the caller to deal with the events one page at a time or `StreamExt::fold` into a vector with all the events as it is today.

PR #291, opened 2020-04-27
[Merged by Bors] - [Events] New Log Filter API

nlordell:
This PR defines a new log stream/log filter API for querying past logs. This is the start of a refactor of the current event querying and streaming implementation in `ethcontract` so that we can have a `eth_getLogs` based log and event stream due to some issues we've encountered with using filters behind a load balancer.

The newly introduced stream is also "fault-tolerant" in that it can deal with a node being temporarily down by retrying the request it was on.

Note that this API is currently unused, but is the start of a larger refactor captured by: #290

### Test Plan

New unit tests.


src/log.rs
```diff
@@ -81,6 +84,376 @@ type CompatFilterStream<T, R> = Compat01As03<FilterStream<T, R>>;
...
+            state: PastLogState::Init { from_block },
+        }
+    }
+}
+
+impl<T: Transport> Stream for PastLogStream<T> {
+    type Item = Result<Vec<Log>, ExecutionError>;
+
+    #[project]
+    fn poll_next(mut self: Pin<&mut Self>, cx: &mut Context) -> Poll<Option<Self::Item>> {
```
e00E: I wonder, could we write this instead as `async fn next() -> Option<Result<Vec<Log, ExecutionError>>` that works like it does for iterators? That way the implementation would be simpler. Then when we have this function we can turn it into a stream in a couple of lines if we still want the stream or there might even be a function the futures crate that already does this.

bh2smith: Interesting observation. I like when things are easily accessible!

src/log.rs
```diff
@@ -81,6 +84,376 @@ type CompatFilterStream<T, R> = Compat01As03<FilterStream<T, R>>;
 /// that can be queried in order to stream logs.
 type CompatCreateFilter<T, R> = Compat01As03<CreateFilter<T, R>>;
 
+/// The default poll interval to use for polling logs from the block chain.
+#[cfg(not(test))]
+pub const DEFAULT_POLL_INTERVAL: Duration = Duration::from_secs(5);
+#[cfg(test)]
+pub const DEFAULT_POLL_INTERVAL: Duration = Duration::from_secs(0);
```
bh2smith, line 91: This seems kinda weird... Shouldn't test related things be declared within the tests?

bh2smith (approved):
Looks good so then we can filter for logs by topic. Does this mean then that we would parallelize our event listener to have multiple services watching for/handling single events? I guess the topic filter can be more than just a single topic.

Look good to me, but would like to hear more about @e00E's suggestion.

src/log.rs
nlordell: This is just saying that the poll interval is 0 when running in test mode, so that futures resolve immediately for unit tests.

src/log.rs
nlordell: Yeah, this PR was from 26 days ago before moving to async/await, I had just forgotten to request a review.

This is a cool idea - I will re-implement with something like [`stream::unfold`](https://docs.rs/futures/0.3.5/futures/stream/fn.unfold.html)!

src/log.rs
```diff
@@ -81,13 +82,316 @@ type CompatFilterStream<T, R> = Compat01As03<FilterStream<T, R>>;
...
+/// stream that constantly queries new logs and deals with re-orgs.
+#[derive(Debug)]
+#[must_use = "log filter builders do nothing unless you query or stream them"]
+pub struct LogFilterBuilder<T: Transport> {
+    /// The underlying web3 provider used for retrieving logs.
+    web3: Web3<T>,
+    /// The block to start streaming logs from.
+    pub from_block: Option<BlockNumber>,
+    /// The block to stop streaming logs from.
+    pub to_block: Option<BlockNumber>,
```
e00E: Mention here what it means when BlockNumber is Latest or Pending.

src/log.rs
```diff
@@ -81,13 +82,316 @@ type CompatFilterStream<T, R> = Compat01As03<FilterStream<T, R>>;
...
+            web3,
+            to_block,
+            block_page_size,
+            filter,
+            page_block,
+            end_block,
+        }
+    }
+
+    async fn next_page(&mut self) -> Result<Option<Vec<Log>>, ExecutionError> {
```
e00E, line 349: Split into multiple smaller functions if sensible. If it doesn't turn out to work well then not needed.

src/log.rs
```diff
@@ -81,13 +82,316 @@ type CompatFilterStream<T, R> = Compat01As03<FilterStream<T, R>>;
...
+            None => {
+                let latest_block = self.web3.eth().block_number().compat().await?.as_u64();
+                self.end_block = Some(latest_block);
+                latest_block
+            }
+        };
+
+        while page_block <= end_block {
+            // NOTE: Log block ranges are inclusive.
+            let page_end = page_block + self.block_page_size - 1;
```
e00E: Maybe assert that block_page_size is > 0 to prevent infinite loop.

src/log.rs
```diff
@@ -81,13 +82,316 @@ type CompatFilterStream<T, R> = Compat01As03<FilterStream<T, R>>;
...
+            } else {
+                // NOTE: The last page is handled a bit differently
+                //   by using the `to_block` that was originally
+                //   specified to the builder. This is done in case
+                //   the to block was "latest" or "pending", where
+                //   we want to make sure that the last call
+                //   includes blocks that have been added since the
+                //   start of the paginated stream.
+                self.to_block
+            };
```
e00E: It would also be sensible to say that the last block is determined when PastLogsParser is created which could make the implementation here simpler in several places.

e00E (approved):

src/log.rs
nlordell: I think I need to refactor this a bit. You are right, things are a little complicated at the moment.

nlordell:
@e00E Let me know what you think of this simpler implementation.

src/log.rs
```diff
@@ -81,13 +84,327 @@ type CompatFilterStream<T, R> = Compat01As03<FilterStream<T, R>>;
...
+    /// This value can be `None` if it `from_block` was specified as `Latest` or
+    /// `Pending`.
+    page_block: u64,
+    /// The last block used for pagination. This is slightly different than
+    /// `to_block` as this must be a concrete block number (and can't be block
+    /// aliases such as `Earliest` or `Latest`).
+    ///
+    /// This has a value of `None` if it has not yet been retrieved (in case
+    /// `to_block` is `Latest` or `Pending`).
+    end_block: u64,
```
e00E: These comments seem out of date because there is no Option anymore.

src/log.rs
```diff
@@ -81,13 +84,327 @@ type CompatFilterStream<T, R> = Compat01As03<FilterStream<T, R>>;
...
+            let page_to_block = if page_end < self.end_block {
+                BlockNumber::Number(page_end.into())
+            } else {
+                // NOTE: The last page is handled a bit differently by using the
+                //   `to_block` that was originally specified to the builder.
+                //   This is done because the `end_block` is determined when the
+                //   pager is created, and it is possible that the `to_block`
+                //   was specified as "latest" or "pending" which may have
+                //   changed since the paging started.
+                self.to_block
```
e00E, line 367: Can't we get rid of this block and always use `self.end_block`. That's what I meant previously that changing the behavior to determine the last block on creation would simplify this part.

```rust
let page_to_block = std::cmp::min(page_end, self.end_block);
```

src/log.rs
```diff
@@ -81,13 +84,327 @@ type CompatFilterStream<T, R> = Compat01As03<FilterStream<T, R>>;
...
+        let next = match (start_block, end_block) {
+            (Some(page_block), Some(end_block)) => PastLogsStream::Paging(PastLogsPager {
+                web3,
+                to_block,
+                block_page_size,
+                filter,
+                page_block,
+                end_block,
+            }),
+            _ => PastLogsStream::Querying(web3, filter.build()),
```
e00E, line 322: What is the point of Querying? Isn't this the same as doing Paging once if we would set `start_block` to the same value as `end_block` above?

src/log.rs
nlordell: Ah, sorry, I misunderstood. The only thing I see with that is that is what does "pending" mean in that case? `latest_block + 1`? Do nodes even allow such a query?

e00E: Your argument makes sense. We can keep it like this.

src/log.rs
nlordell: This branch is required for when we can't determine the initial `page_block` or `end_block`, and want to do a single query to the node with the provided `to_block` and `from_block`.

This mostly stems from the fact that you can specify some things that don't really make sense range wise, for example: `filter.from_block(BlockNumber::Pending).to_block(BlockNumber::Earliest)`. What does this query even mean? Instead of coming up with errors, I just make a query and differ the decision on how to handle this case to the node.

e00E: I see. Could also argue that we shouldn't allow this and get rid of more code but I don't mind either way.

nlordell: I'll keep it as-is so it works like the `query_past_logs_paginated` used to. It also short-circuits the case where you do:
```
LogFilterBuilder::new(web3).past_log_pages();
// equivalent to
LogFilterBuilder::new(web3).from_block(Latest).to_block(Latest).past_log_pages();
```

Which should work and not require fetching the `end_block`.

bh2smith:
> bors changed the title [Events] New Log Filter API [Merged by Bors] - [Events] New Log Filter API

 I do not like the way this looks. Are we really gonna use this thing?

fleupold:
Has this been merged? It's not really clear from the status (it just says closed)

nlordell:
Yeah, this is what @bh2smith was talking about where it marks the PR as closed and renames it to "[Merged by Bors] <original PR name>".

Here is the full discussion - including the reason why the PR gets closed this way: https://forum.bors.tech/t/use-a-squash-merge-in-bors/349

fleupold:
Would be nice if it could at least link to the Commit that was merged.

nlordell:
It does already: 
image

fleupold:
> It does already:

I'm blind. Can we also get it to delete branches with closed PRs?

PR #293, opened 2020-04-29
Fallback to polling on filter query error

nlordell:
Fallback to polling on filter query error. This makes confirmation polling a bit more robust in cases where nodes are sitting behind load balancers or filters get unexpectedly cleaned up.

### Test Plan

New unit test.


e00E (approved):

fleupold (approved):
Nice, I think it's a sign of well designed code that this change fits in so nicely.

bh2smith (approved):

PR #304, opened 2020-05-19
[Async] Migrate `Contract::deployed` to use async/await

nlordell:
This PR migrates `Contract::deployed` to use async/await. Note that a lot of code gets removed because of this since we no longer need to manually implement futures and chaining futures together is much easier.

Note that the manual deployments now get injected into the runtime artifact instance.

### Test Plan

Migrated unit tests, and example remained unchanged (so no major API usage changes) and will get tested on CI.


generate/src/contract/common.rs
```diff
@@ -136,3 +161,40 @@ pub(crate) fn expand(cx: &Context) -> TokenStream {
         }
     }
 }
+
```
nlordell, line 164: Note that this code was moved from `generate/src/contract/deployment.rs` to here.

e00E (approved):

PR #308, opened 2020-05-22
[Merged by Bors] - Euclidean Division and Remainder

### LINKED ISSUE #253: Implement division for `I256`

bh2smith:
Note that, for consistency, a few changes were made to the non-euclidean analogue.

- [x] div_euclid
- [x] rem_euclid
- [x] checked_div_euclid
- [x] overflowing_div_euclid
- [x] wrapping_div_euclid
- [x] overflowing_rem_euclid
- [x] wrapping_rem_euclid
- [x] checked_rem_euclid

Closes #253

### DIFF
--- a/src/int.rs
+++ b/src/int.rs
@@ -717,6 +717,22 @@ impl I256 {
         self.overflowing_rem(rhs).0
     }
 
+    /// Calculates the quotient of Euclidean division of self by rhs.
+    /// This computes the integer n such that self = n * rhs + self.rem_euclid(rhs),
+    /// with 0 <= self.rem_euclid(rhs) < rhs.
+    /// In other words, the result is self / rhs rounded to the integer n such that self >= n * rhs.
+    /// If self > 0, this is equal to round towards zero (the default in Rust);
+    /// if self < 0, this is equal to round towards +/- infinity.
+    pub fn div_euclid(self, rhs: Self) -> Self {
+        let q = self / rhs;
+        if (self % rhs).is_negative() {
+            return if rhs.is_positive() { q - I256::one() } else { q + I256::one() }
+        }
+        q
+    }
+
+
+
     /// Returns the sign of `self` to the exponent `exp`.
     ///
     /// Note that this method does not actually try to compute the `self` to the
@@ -1486,6 +1502,31 @@ mod tests {
     fn division_by_zero() {
         let _ = I256::one() / I256::zero();
     }
+
+    #[test]
+    fn div_euclid() {
+        let a = I256::from(7);
+        let b = I256::from(4);
+
+        assert_eq!(a.div_euclid(b), I256::one()); // 7 >= 4 * 1
+        assert_eq!(a.div_euclid(-b), -I256::one()); // 7 >= -4 * -1
+        assert_eq!((-a).div_euclid(b), -I256::from(2)); // -7 >= 4 * -2
+        assert_eq!((-a).div_euclid(-b), I256::from(2)); // -7 >= -4 * 2
+    }
+
+    #[test]
+    #[should_panic]
+    fn div_euclid_by_zero() {
+        let _ = I256::one().div_euclid(I256::zero());
+        assert_eq!(I256::MIN.div_euclid(-I256::one()), I256::MAX);
+    }
+
+    #[test]
+    #[should_panic]
+    fn div_euclid_overflow() {
+        I256::MIN.div_euclid(-I256::one());
+    }
+
 
     #[test]
     #[should_panic]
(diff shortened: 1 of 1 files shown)


src/int.rs
```diff
@@ -717,6 +715,104 @@ impl I256 {
         self.overflowing_rem(rhs).0
     }
 
+    /// Calculates the quotient of Euclidean division of self by rhs.
+    /// This computes the integer n such that self = n * rhs + self.rem_euclid(rhs),
+    /// with 0 <= self.rem_euclid(rhs) < rhs.
+    /// In other words, the result is self / rhs rounded to the integer n such that self >= n * rhs.
+    /// If self > 0, this is equal to round towards zero (the default in Rust);
+    /// if self < 0, this is equal to round towards +/- infinity.
```
nlordell: nit: Doc comments in Rust are markdown, so this will be formatted like so:

> Calculates the quotient of Euclidean division of self by rhs. This computes the integer n such that self = n * rhs + self.rem_euclid(rhs), with 0 <= self.rem_euclid(rhs) < rhs. In other words, the result is self / rhs rounded to the integer n such that self >= n * rhs. If self > 0, this is equal to round towards zero (the default in Rust); if self < 0, this is equal to round towards +/- infinity.

```suggestion
    /// Calculates the quotient of Euclidean division of self by rhs.
    ///
    /// This computes the integer `n` such that `self = n * rhs + self.rem_euclid(rhs)`,
    /// with `0 <= self.rem_euclid(rhs) < rhs`.
    /// In other words, the result is `self / rhs` rounded to the integer `n` such that `self >= n * rhs`:
    /// * If `self > 0`, this is equal to round towards zero (the default in Rust);
    /// * If `self < 0`, this is equal to round towards +/- infinity.
```

Will have nicer formatting:
> Calculates the quotient of Euclidean division of self by rhs.
>
> This computes the integer `n` such that `self = n * rhs + self.rem_euclid(rhs)`,
> with `0 <= self.rem_euclid(rhs) < rhs`.
> In other words, the result is `self / rhs` rounded to the integer `n` such that `self >= n * rhs`:
> * If `self > 0`, this is equal to round towards zero (the default in Rust);
> * If `self < 0`, this is equal to round towards +/- infinity.

Same for other comments.

nlordell (approved):
Awesome! Just some issues with the doc-comment formatting.

nlordell:
bors try

bh2smith:
> bors try

WAT?

nlordell:
>  > bors try
>
> WAT?

Its how you trigger the CI. It failed because of formatting.

nlordell:
Here are the bors commands: https://bors.tech/documentation/

bh2smith:
So what, like do I have to ask bors to run every time I want it?

bh2smith:
bors try

bh2smith:
bors merge

fedgiac (commented):
Everything looks right to me.

To nitpick, I see that the function `div_euclid` and `rem_euclid` use basically the same logic but have different implementations. Maybe they could be made more similar?
Even better, if the compiler is clever enough to optimize away variables that aren't used, then there could be a single implementation `dif_rem_euclid` returning a pair struct, which could be useful by itself, and each of `div_euclid` and `rem_euclid` would refer to their corresponding pair element with no loss in efficiency.

bh2smith:
I agree that something like `divmod` would probably make more sense, but in this case I was trying to stick as close as possible to the existing rust documentation. That is, I was trying to mimic what is already part of the standard library for `i32` found here in https://doc.rust-lang.org/std/primitive.i32.html


Python (on the other hand) implements what you suggest already via a function called [divmod](https://docs.python.org/3/library/functions.html#divmod). Here is my terminal output
```py
➜  ~ python3
Python 3.7.3 (default, Apr 24 2020, 18:51:23) 
[Clang 11.0.3 (clang-1103.0.32.62)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> divmod(5, 2)
(2, 1)
>>> divmod(-5, -2)
(2, -1)
>>> divmod(-5, 2)
(-3, 1)
>>> divmod(5, -2)
(-3, -1)
```

PR #309, opened 2020-05-22
[Merged by Bors] - [Async] Implement event querying with new log API

nlordell:
This PR changes the query methods on the `AllEventsBuilder` type to be async, making use of the new `LogFilterBuilder` API.

### Test Plan

Unit tests still pass.


src/contract/event.rs
```diff
@@ -247,53 +247,29 @@ impl<T: Transport, E: Detokenize> Stream for EventStream<T, E> {
...
-    /// does not allow access to these values once stored.
-    pub from_block: Option<BlockNumber>,
-    /// The last block to retrieve logs for.
-    pub to_block: Option<BlockNumber>,
-    /// The topic filters that are encoded based on the event ABI.
-    pub topics: TopicFilter,
-    /// The polling interval for querying the node for more events.
-    pub poll_interval: Option<Duration>,
+    /// The underlying log filter for these contract events.
+    pub filter: LogFilterBuilder<T>,
```
bh2smith, line 255: By the looks of it, all the removed attributes are now part of `FilterBuilder`. Is this correct?

src/contract/event.rs
```diff
@@ -329,60 +297,56 @@ impl<T: Transport, E: ParseLog> AllEventsBuilder<T, E> {
...
-        self.poll_interval = Some(value);
+        self.filter = self.filter.topic3(topic);
         self
     }
 
     /// The page size in blocks to use when doing a paginated query on past
     /// events.
     pub fn block_page_size(mut self, value: u64) -> Self {
-        self.block_page_size = Some(value);
+        self.filter = self.filter.block_page_size(value);
```
bh2smith, line 325: This seems incredibly weird to me...

Like why not just `self.filter.block_page_size = value`?

Is this some sort of meta recursion?

bh2smith (approved):
Looks cool, but there is some fishy meta-recursion going on in some places that I have pointed out. Wondering whats up with that.

src/contract/event.rs
nlordell: This is because `LogFilterBuilder::block_page_size` takes in a `LogFilterBuilder` as `self` and returns an updated `LogFilterBuilder`. Also, `self.filter.block_page_size(value)` does some value checking (for example panicking if `value == 0`).

e00E (approved):

nlordell:
bors retry

PR #315, opened 2020-05-25
[Merged by Bors] - [Events] Add `limit` option back

nlordell:
The limit option was removed during a refactor, this PR adds it back.

### Test Plan

New unit test to make sure it is properly being added to the logs query.

### DIFF
--- a/src/contract/event.rs
+++ b/src/contract/event.rs
@@ -96,6 +96,14 @@ impl<T: Transport, E: Detokenize> EventBuilder<T, E> {
         self
     }
 
+    /// Limit the number of events that can be retrieved by this filter.
+    ///
+    /// Note that this parameter is non-standard.
+    pub fn limit(mut self, value: usize) -> Self {
+        self.filter = self.filter.limit(value);
+        self
+    }
+
     /// The polling interval. This is used as the interval between consecutive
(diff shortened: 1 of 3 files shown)


e00E (approved):

src/contract/event.rs
```diff
@@ -229,6 +237,14 @@ impl<T: Transport, E: ParseLog> AllEventsBuilder<T, E> {
         self
     }
 
+    /// Limit the number of events that can be retrieved by this filter.
+    ///
+    /// Note that this parameter is non-standard.
+    pub fn limit(mut self, value: usize) -> Self {
+        self.filter = self.filter.limit(value);
+        self
+    }
```
bh2smith, line 246: I see this snippet in more than 1 place... Is this intended?

bh2smith (approved):
Looks good. I only wonder why was it ever removed in the first place?

src/contract/event.rs
nlordell: > I only wonder why was it ever removed in the first place?

It was removed as part of a refactor, with intention of adding it back later.

> Is this intended?

Yes, there are more than one kind of builder that can set the `limit` property. These methods are just small helpers for setting that property.

PR #325, opened 2020-05-26
[WIP/Async] Wait for target block

nlordell:
This PR modifies the behaviour of transaction confirmations by making it so the filter is re-queried in case the target block is not reached (in case there is a reorg while waiting for blocks to be mined).

Now that I think about this more (after implementing it :facepalm:) - I think it should stay as is. The reason being if there is a reorg, then the transaction receipt might have been invalidated so its worth re-querying to re-calculate the remaining blocks that need to be waited for anyway (the transaction might even have moved up a block for example).

Just wanted to share the implementation in case you guys think this should be included - otherwise I think the PR should be closed and confirmations should be kept as is.

### DIFF
--- a/src/transaction/confirm.rs
+++ b/src/transaction/confirm.rs
@@ -11,7 +11,7 @@ use futures::compat::Future01CompatExt;
 use futures_timer::Delay;
 use std::time::Duration;
 use web3::api::Web3;
-use web3::futures::Stream as Stream01;
+use web3::futures::Stream as _;
 use web3::types::{TransactionReceipt, H256, U64};
 use web3::Transport;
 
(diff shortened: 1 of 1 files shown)


fleupold (commented):
If I understand correctly we are still requerying the tx receipt on each check, we just don't double check the current block number (which even in case of a reorg shouldn't have changed).

I dont fully understand your concern here?

nlordell:
I added a (hopefully) better explanation to the other PR.

nlordell:
Closing this, see discussion https://github.com/gnosis/ethcontract-rs/pull/323#discussion_r430306987

PR #340, opened 2020-06-30
Update web3

e00E:
Draft because the current web3 release is a pre-release and there are some bugs that are fixed on master but not in that release.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -32,7 +32,7 @@ members = [
 [dependencies]
 ethcontract-common = { version = "0.7.2", path = "./common" }
 ethcontract-derive = { version = "0.7.2", path = "./derive", optional = true}
-futures = { version = "0.3", features = ["compat"] }
+futures = "0.3"
 futures-timer = "3.0"
 hex = "0.4"
 jsonrpc-core = "14.0"
@@ -42,5 +42,8 @@ serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
 thiserror = "1.0"
 uint = "0.8"
-web3 = "0.11"
+web3 = "0.13"
 zeroize = "1.1"
+
+[dev-dependencies]
+tokio = { version = "0.2", features = ["macros"] }
--- a/examples/sources.rs
+++ b/examples/sources.rs
@@ -7,18 +7,14 @@ ethcontract::contract!(
 );
 ethcontract::contract!("npm:@gnosis.pm/owl-token@3.1.0/build/contracts/TokenOWL.json");
 
-fn main() {
-    futures::executor::block_on(run());
-}
-
-async fn run() {
+#[tokio::main]
+async fn main() {
     let infura_url = {
         let project_id = env::var("INFURA_PROJECT_ID").expect("INFURA_PROJECT_ID is not set");
         format!("https://mainnet.infura.io/v3/{}", project_id)
     };
 
-    let (eloop, http) = Http::new(&infura_url).expect("transport failed");
-    eloop.into_remote();
+    let http = Http::new(&infura_url).expect("transport failed");
     let web3 = Web3::new(http);
 
     let instance = Owl::deployed(&web3)
--- a/generate/src/contract/common.rs
+++ b/generate/src/contract/common.rs
@@ -68,10 +68,9 @@ pub(crate) fn expand(cx: &Context) -> TokenStream {
                 address: self::ethcontract::Address,
             ) -> Self
             where
-                F: self::ethcontract::web3::futures::Future<
-                    Item = self::ethcontract::json::Value,
-                    Error = self::ethcontract::web3::Error,
-                > + Send + 'static,
+                F: std::future::Future<
+                       Output = Result<self::ethcontract::json::Value, self::ethcontract::web3::Error>
+                   > + Send + Unpin + 'static,
                 T: self::ethcontract::web3::Transport<Out = F> + Send + Sync + 'static,
             {
                 Contract::with_transaction(web3, address, None)
@@ -92,10 +91,9 @@ pub(crate) fn expand(cx: &Context) -> TokenStream {
                 transaction_hash: Option<self::ethcontract::H256>,
             ) -> Self
             where
-                F: self::ethcontract::web3::futures::Future<
-                    Item = self::ethcontract::json::Value,
-                    Error = self::ethcontract::web3::Error,
-                > + Send + 'static,
+                F: std::future::Future<
+                       Output = Result<self::ethcontract::json::Value, self::ethcontract::web3::Error>
+                   > + Send + Unpin + 'static,
                 T: self::ethcontract::web3::Transport<Out = F> + Send + Sync + 'static,
             {
                 use self::ethcontract::Instance;
--- a/src/contract/event.rs
+++ b/src/contract/event.rs
@@ -8,7 +8,6 @@ use crate::errors::{EventError, ExecutionError};
 use crate::log::LogFilterBuilder;
 pub use ethcontract_common::abi::Topic;
 use ethcontract_common::abi::{Event as AbiEvent, RawTopicFilter, Token};
-use futures::compat::Future01CompatExt as _;
 use futures::future::{self, TryFutureExt as _};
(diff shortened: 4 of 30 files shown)


examples/sources.rs
```diff
@@ -1,3 +1,6 @@
+fn main() {}
+
+/*
```
e00E: TODO: revert this before merging
I comment this out because I get compilation errors in my editor because etherscan rate limits me.

src/contract/event.rs
```diff
@@ -446,7 +444,7 @@ mod tests {
             .stream()
             .boxed()
             .next()
-            .immediate()
+            .wait()
```
e00E, line 447: `immediate` no longer works probably because of internals in web3 changed.

src/lib.rs
```diff
@@ -125,7 +125,7 @@ pub mod prelude {
     pub use crate::transaction::{Account, GasPrice};
     pub use web3::api::Web3;
     pub use web3::transports::Http;
-    pub use web3::types::{Address, BlockNumber, TransactionCondition, H160, H256, U256};
+    pub use web3::types::{Address, BlockId, BlockNumber, TransactionCondition, H160, H256, U256};
```
e00E, line 128: This is needed for a few functions in web3 so we probably want to reexport it.

src/transaction/confirm.rs
```diff
@@ -198,19 +192,14 @@ impl<T: Transport> ConfirmationContext<'_, T> {
...
-            .await
-            .map_err(|(err, _)| err)?;
+        let block_filter = self.web3.eth_filter().create_blocks_filter().await?;
+        let mut stream = block_filter.stream(self.params.poll_interval);
+        for _ in 0..block_count {
+            stream
+                .next()
+                .await
+                .ok_or(ExecutionError::StreamEndedUnexpectedly)??;
+        }
```
e00E, line 208: By turning `skip` into a loop we make sure to notice observe all errors.

Cargo.toml
```diff
@@ -42,5 +42,8 @@ serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
 thiserror = "1.0"
 uint = "0.8"
-web3 = "0.11"
+web3 = "0.12"
 zeroize = "1.1"
+
+[dev-dependencies]
+tokio = { version = "0.2", features = ["macros"] }
```
e00E, line 49: The web3 provided http transport only runs on the Tokio runtime so for our examples we need this.

examples/sources.rs
nlordell: You can log in to etherscan and setup an API key :smile:

generate/src/contract/common.rs
```diff
@@ -68,10 +68,9 @@ pub(crate) fn expand(cx: &Context) -> TokenStream {
                 address: self::ethcontract::Address,
             ) -> Self
             where
-                F: self::ethcontract::web3::futures::Future<
-                    Item = self::ethcontract::json::Value,
-                    Error = self::ethcontract::web3::Error,
-                > + Send + 'static,
+                F: std::future::Future<
+                       Output = Result<self::ethcontract::json::Value, self::ethcontract::web3::Error>
+                   > + Send + Unpin + 'static,
```
nlordell, line 73: I just noticed that `web3::Transport::Future: Unpin` :open_mouth:

src/contract/event.rs
nlordell: Hmm, is it just for this method? It would be nice to know what `web3` method exactly is always yielding.

src/contract/method.rs
```diff
@@ -264,15 +263,14 @@ impl<T: Transport, R: Detokenizable> ViewMethodBuilder<T, R> {
             .call(
                 CallRequest {
                     from: self.m.tx.from.map(|account| account.address()),
-                    to: self.m.tx.to.unwrap_or_default(),
+                    to: Some(self.m.tx.to.unwrap_or_default()),
```
nlordell, line 266: :tada: We can get rid of the `estimate_gas` hack!

src/errors.rs
```diff
@@ -99,6 +99,10 @@ pub enum ExecutionError {
     /// A removed log was received when querying past logs.
     #[error("unexepected removed log when querying past logs")]
     RemovedLog(Box<Log>),
+
+    /// A stream ended unexpectedly.
+    #[error("stream ended unexpectedly")]
```
nlordell: nit: maybe indicate its a log stream in the error?

src/log.rs
```diff
@@ -515,7 +511,7 @@ mod tests {
             .stream()
             .boxed()
             .next()
-            .immediate()
+            .wait()
```
nlordell, line 514: Ah OK, it appears to be related to the log stream! I think its not really an important issue then.

src/test/prelude.rs
```diff
@@ -18,3 +18,15 @@ impl<F: Future + Sized> FutureImmediateExt for F {
             .expect("future did not resolve immediately")
     }
 }
+
+/// An extension future to wait for a future.
+pub trait FutureWaitExt: Future {
```
nlordell: Maybe the `wait` and `immediate` methods can be combined into a `FutureTestExt` trait instead of having two separate traits for testing.

src/transaction/confirm.rs
```diff
@@ -298,7 +287,7 @@ mod tests {
         transport.add_response(generate_tx_receipt(hash, 2));
 
         let confirm = wait_for_confirmation(&web3, hash, ConfirmParams::mined())
-            .immediate()
+            .wait()
```
nlordell, line 296: I see now, this is probably related to how `Stream`s are implemented in `web3` that make it always yield. Maybe related to timers?

src/transaction/estimate_gas.rs
```diff
@@ -3,7 +3,6 @@
```
nlordell, line 3: I suspect that the `estimate_gas` module can be removed now that the `to` parameter is optional. I'd also be OK if you prefer to do it in a follow up PR.

src/transport.rs
```diff
@@ -6,18 +6,20 @@
...
+use std::future::Future;
 use std::sync::Arc;
 use web3::error::Error as Web3Error;
-use web3::futures::Future;
 use web3::{RequestId, Transport};
 
 /// Type alias for the output future in for the `DynTransport`'s `Transport`
 /// implementation.
-type BoxedFuture = Box<dyn Future<Item = Value, Error = Web3Error> + Send + 'static>;
+type BoxedFuture = BoxFuture<'static, Result<Value, Web3Error>>;
```
nlordell, line 22: Yay! Futures 0.3 :tada:

nlordell (approved):
I think it looks great already!

src/transaction/confirm.rs
nlordell: There is also `TryStreamExt::try_skip_while` (but no `try_skip` :confused:) but I think the `for .. in` loop is elegant here.

e00E:
@nlordell I think this is ready. With the current web3 version all examples work.

examples/sources.rs
```diff
@@ -7,18 +10,14 @@ ethcontract::contract!(
 );
 ethcontract::contract!("npm:@gnosis.pm/owl-token@3.1.0/build/contracts/TokenOWL.json");
 
-fn main() {
-    futures::executor::block_on(run());
-}
-
-async fn run() {
+#[tokio::main]
```
nlordell, line 10: Don't forget to un-comment this before merging.

src/test/prelude.rs
```diff
@@ -6,13 +6,18 @@ pub use serde_json::json;
 use std::future::Future;
 pub use web3::api::Web3;
 
-/// An extension future to assert that a future resolves immediately.
-pub trait FutureImmediateExt: Future {
+/// An extension future to wait for a future.
+pub trait FutureTestExt: Future {
+    /// Block thread on a future completing.
+    fn wait(self) -> Self::Output;
     /// Block thread on a future completing.
```
nlordell: Just realized this doc comment is wrong from before, so you mind changing it here already?

nlordell (approved):
Looks good.

PR #359, opened 2020-07-31
Make web3 features optional

e00E:
These web3 features introduce more dependencies which should be optional
for users of ethcontract.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,9 +17,13 @@ to Ethereum smart contracts.
 name = "ethcontract"
 
 [features]
-default = ["derive"]
+default = ["derive", "http", "http-tls", "ws", "ws-tls"]
 derive = ["ethcontract-derive"]
 samples = []
+http = ["web3/http"]
+http-tls = ["web3/http-tls"]
+ws = ["web3/ws"]
+ws-tls = ["web3/ws-tls"]
 
 [workspace]
 members = [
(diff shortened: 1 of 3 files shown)


src/lib.rs
```diff
@@ -124,6 +124,7 @@ pub mod prelude {
     pub use crate::secret::{Password, PrivateKey};
     pub use crate::transaction::{Account, GasPrice};
     pub use web3::api::Web3;
+    #[cfg(feature = "http")]
     pub use web3::transports::Http;
```
e00E, line 128: Should we remove this re-export alltogether?

Cargo.toml
```diff
@@ -17,9 +17,13 @@ to Ethereum smart contracts.
 name = "ethcontract"
 
 [features]
-default = ["derive"]
+default = ["derive", "http", "http-tls", "ws", "ws-tls"]
```
e00E, line 20: I also tried not changing the default features but then setting up the examples becomes more difficult.
Most examples need "http" but I haven't found a good way to enable an ethcontract feature for the examples. If these features aren't default then the examples need to be run like `cargo run --example abi --feature http`.
tokio does it by having the examples by it's own crate that can depend on `../` with all features enabled.

nlordell: I think its fine to have these enabled by default. I think cargo does not allow you to specify features for examples only, you can however specify `required-features` so that the example doesn't run unless the features are specified on the command line, but this becomes combersome.

src/lib.rs
nlordell: Hmm, I think this is the most commonly used transport, so its nice to have it in the `prelude`. But I'm also don't have a strong opinion.

nlordell (approved):
Nice! Great feature.

PR #395, opened 2020-11-04
Fix clippy warnings and CI

e00E:
For the CI part I do not know what actually broke. What we observe is that the examples can't connect to truffle when run through `yarn start`. This is fixed by running `truffle develop` directly but `npm install -g truffle` fails on `ubuntu:bionic` so I upgraded to `ubuntu:focal`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,15 +3,15 @@ branches:
     - main
 
 os: linux
-dist: bionic
+dist: focal
 language: rust
 
 jobs:
   fast_finish: true
   include:
     - stage: "Build and Test"
-      name: "Rust: 1.42.0"
-      rust: 1.42.0
+      name: "Rust: 1.47.0"
+      rust: 1.47.0
     - name: "Rust: stable"
       rust: stable
     - name: "Rust: beta"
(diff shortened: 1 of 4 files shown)


nlordell (approved):

.travis.yml
```diff
@@ -22,10 +22,10 @@ jobs:
...
       rust: stable
       script:
-        # NOTE: We pipe `tail -f /dev/null` into `yarn start`, which internally
-        #   calls `truffle develop`, so that the process' STDIN stays open, as
-        #   `truffle develop` exits as soon as it gets an EOF from STDIN.
-        - tail -f /dev/null | yarn --cwd examples/truffle start > /dev/null &
+        - npm install -g truffle
+        # NOTE: We pipe `tail -f /dev/null`, so that the process' STDIN stays
+        # open, as `truffle develop` exits as soon as it gets an EOF from STDIN.
+        - (cd examples/truffle; tail -f /dev/null | truffle develop > /dev/null) &
```
nlordell, line 28: What if you use `npx truffle` here? I'm afraid this build is quite brittle as it newer versions of truffle can break it un unexpected ways.
```suggestion
        - (cd examples/truffle; tail -f /dev/null | npx truffle develop > /dev/null) &
```

nlordell: If that doesn't work, you can
- Update to latest version of Truffle (the Ubuntu bionic thing might be why #392 is failing?)
- Even run `truffle` directly via `./node_modules/.bin/truffle`.

e00E: `./node_modules/.bin/truffle` does not work (it seems to hang forever). i'm guessing `yarn start` and `npx truffle` call that same binary which has some issue. With `strace` I see that it seems to be waiting on the unix socket `/tmp/truffle.develop` but I don't know what that means.
I am going to merge this PR now so that CI is at least fixed now even though it could be brittle in the future. Will make issue about this though.

PR #397, opened 2020-11-07
test int: div_euclid_overflow; handle panic

Horki:
-

### DIFF
--- a/src/int.rs
+++ b/src/int.rs
@@ -8,6 +8,7 @@ use std::convert::{TryFrom, TryInto};
 use std::fmt;
 use std::iter;
 use std::ops;
+use std::panic;
 use std::str;
 use std::{i128, i64, u64};
 use web3::contract::{self, tokens};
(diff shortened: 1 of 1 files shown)


nlordell:
Thank's for the contribution and for finding an issue! Indeed this test fails when building without `debug_assertions`. I would much rather prefer if the panic wasn't manually caught using `catch_unwind` but instead the `should_panic` attribute.

So, right now if I:
```
$ cargo test div_euclid_overflow          
    Finished test [unoptimized + debuginfo] target(s) in 0.08s
     Running target/debug/deps/ethcontract-040189939399e07f

running 1 test
test int::tests::div_euclid_overflow ... ok
```

However, in release mode,
```
$ cargo test div_euclid_overflow --release
    Finished release [optimized] target(s) in 0.09s
     Running target/release/deps/ethcontract-9c4ae386f5d49c66

running 1 test
test int::tests::div_euclid_overflow ... FAILED
```

I think the easiest solution would be to make the `should_panic` attribute in the unit test conditional on a debug build. Something like
```rust
#[test]
#[cfg_attr(debug_assertions, should_panic)]
fn div_euclid_overflow_in_debug() {
    I256::MIN.div_euclid(-I256::one());
}
```

PR #425, opened 2020-12-19
update web3 and jsonrpc core

### LINKED ISSUE #426: Update web3 to 0.14
I added this since I'm using both of these packages and I wanted to upgrade on mine project. (:
Thankfully the update was straight forward enough and one attempt at it can be seen at #425

lsunsi:
### What
I tried updating the web3 and jsonrpc core (required by both web3 and this library).
I relied on tests and compilation errors to know if we're good or not haha.

### Considerations
I tried adding both tokio and async-std new feature flags into this package, but it failed building due to the 'all features' check. Only tokio or async-std should be enabled one at a time, so I made the choice for tokio (like it was before).

### Changes
- Update web3 (to 0.14) and jsonrpc-core (to 0.15)
- Fix compilation issue on wait_for_blocks_with_filter

Closes #426

### DIFF
--- a/src/transaction/confirm.rs
+++ b/src/transaction/confirm.rs
@@ -199,7 +199,9 @@ impl<T: Transport> ConfirmationContext<'_, T> {
     /// Waits for a certain number of blocks to be mined using a block filter.
     async fn wait_for_blocks_with_filter(&self, block_count: usize) -> Result<(), ExecutionError> {
         let block_filter = self.web3.eth_filter().create_blocks_filter().await?;
-        let mut stream = block_filter.stream(self.params.poll_interval);
+        let stream = block_filter.stream(self.params.poll_interval);
+        futures::pin_mut!(stream);
+
         for _ in 0..block_count {
             stream
                 .next()
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,13 +17,13 @@ to Ethereum smart contracts.
 name = "ethcontract"
 
 [features]
-default = ["derive", "http", "http-tls", "ws", "ws-tls"]
+default = ["derive", "http", "http-tls", "ws-tokio", "ws-tls-tokio"]
 derive = ["ethcontract-derive"]
 samples = []
 http = ["web3/http"]
 http-tls = ["web3/http-tls"]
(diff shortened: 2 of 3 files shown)


src/transaction/confirm.rs
```diff
@@ -199,12 +199,12 @@ impl<T: Transport> ConfirmationContext<'_, T> {
...
-                .next()
-                .await
-                .ok_or(ExecutionError::StreamEndedUnexpectedly)??;
+        let stream = block_filter.stream(self.params.poll_interval);
+        let results = stream.take(block_count).collect::<Vec<_>>().await;
+        let blocks = results.into_iter().collect::<Result<Vec<_>, _>>()?;
+
+        if blocks.len() < block_count {
+            return Err(ExecutionError::StreamEndedUnexpectedly);
         }
```
e00E: This should be
```suggestion
        let stream = block_filter.stream(self.params.poll_interval);
        futures::pin_mut!(stream);
        for _ in 0..block_count {
            stream
                .next()
                .await
                .ok_or(ExecutionError::StreamEndedUnexpectedly)??;
        }
```
The only new line compared to master is `futures::pin_mut!(stream);`.
The reason we need this is that internally web3 no longer guarantees that the stream is Unpin.

e00E (commented):
Thanks for the PR. I would like to not change the logic of the stream as seen in my suggestion. Other than that it looks good.

src/transaction/confirm.rs
lsunsi: You are right. I was trying to remove this requirement reworking how to work was being done. But I just reverted the changes and added the pin mut like you suggested! (: PR updated

e00E (approved):

e00E:
Note to self: This PR can be merged. Ci fails because two tests (rinkeby, infura) use secrets which aren't available in PRs coming from forks to prevent leaking them.

lsunsi:
@e00E Awesome! If is there anything else you can just ask :D

nlordell (approved):
Looks good to me, thank you so much for your PR! @e00E should we just modify CI to skip tests that require secrets when its a PR from a foreign repo?

PR #429, opened 2020-12-22
Run tests with Github actions

e00E:
Deployments are still handled by travis because we need to move the
secrets to github.
Likewise some examples need secrets to run. We detect this in the script
so that those are only run when available. This is useful for pull
requests from forks because those do not have access to secrets.
Also made a small code change to fix a nightly clippy warnings.


.github/workflows/pull-request.yaml
```diff
@@ -0,0 +1,60 @@
...
+          toolchain: ${{ matrix.rust }}
+          profile: minimal
+          components: rustfmt, clippy
+          default: true
+      - uses: Swatinem/rust-cache@v1
+      - run: cargo fmt -- --check
+      - run: cd examples/truffle && yarn --frozen-lockfile && yarn build
+      - run: cargo clippy --all-features --all-targets -- -D warnings
+      - run: cargo test --all-features
+      - run: |
```
nlordell, line 41: Is it possible to run examples as this as a completely separate "job" so that it is run in parallel?

nlordell (approved):
Nice!

.github/workflows/pull-request.yaml
e00E: It is probably possible but I'm not sure how to do this in a nice way. The whole job is so fast (2 or 3 minutes) anyway that it doesn't matter much. But agree it's something we can keep in mind for the future if we want to optimize it more.

e00E:
Also updated branch protection rules in the repo to use the actions.

PR #430, opened 2020-12-22
Restructure crate to have virtual workspace at the top level

e00E:
According to the cargo documentation
https://doc.rust-lang.org/cargo/reference/workspaces.html the top level
Cargo.toml can include both the crate and the workspace or just the
workspace which is called virtual workspace. I find that less confusing
because we publish several packages from the workspace and not just one
primary.
The new structure has been taken from the tokio-rs/tracing repository.
The ethcontract crate has been moved into a folder, the crate folders
have renamed to the exact crate name and the examples have been turned
into a standalone crate.


.travis.yml
```diff
@@ -25,14 +25,14 @@ jobs:
...
+        - cargo run --package examples --example async
+        - cargo run --package examples --example deployments
+        - cargo run --package examples --example events
         - cargo run --package examples-generate
-        - cargo run --example linked
-        - cargo run --example rinkeby
-        - cargo run --example sources
+        - cargo run --package examples --example linked
+        - cargo run --package examples --example rinkeby
+        - cargo run --package examples --example sources
```
nlordell: I really like moving all the examples to the `examples` crate!

README.md
```diff
@@ -166,16 +166,12 @@ cargo run --example past_events
...
-generation**. In order to view the documentation for these contracts you need to
-first build the contracts and then generate documentation for the crate with the
-`samples` feature enabled:
+You can view example generated contract documentation by fist building the
+contracts and then generating documentation for the crate:
 
 ```sh
 (cd examples/truffle; yarn && yarn build)
-cargo doc --features samples --open
+cargo doc --package examples-documentation --no-deps --open
```
nlordell, line 174: This is definitely the way to do this!

examples/documentation/Cargo.toml
```diff
@@ -0,0 +1,8 @@
+[package]
+name = "examples-documentation"
+version = "0.0.0"
+publish = false
+edition = "2018"
```
nlordell, line 5: nit: missing license.

nlordell (approved):
While personally I prefer `ethcontract` as the root crate, it does seem to be the standard practice for repos with multiple interdependant crates to have a virtual workspace so I think its probably better to do it like this.

The other changes (like examples and example documentation crate) are all awesome!

One nit, could you fix up `ci/bump.sh` in order to not modify the versions in the `xample*` crates?
https://github.com/gnosis/ethcontract-rs/blob/25e0c535e1f07e7e7288571f2b3852ffb1908969/ci/bump.sh#L66

I think it should be changed to something like:
```sh
for manifest in ethcontract*/Cargo.toml; do
```

examples/documentation/Cargo.toml
e00E: I think the license is intended to be used to by crates io to automatically know the license of all packages. We never publish this one so it's not needed.

nlordell: We have licenses for the other `publish = false` crates so maybe we can either add it here as well, or also remove it in those places. My main argument of adding it is that, even if we don't publish this crate, it can still be added as a dependency by the Github repo URL, in which case having a license field may be useful (although, in practice I never see this happening).

PR #450, opened 2021-01-13
Relase to crates.io from github actions

vkgnosis:
I would like to merge this and create a temporary release that we can delete afterwards to test it. If it looks good make another PR to remove the --dry-run.


ci/deploy.sh
```diff
@@ -2,8 +2,11 @@
 
 set -e
 
+# Remove this after testing that it has the correct value.
+echo $GITHUB_REF
```
nlordell: From the [documentation,](https://docs.github.com/en/free-pro-team@latest/actions/reference/environment-variables#default-environment-variables) I think this will have the value:
```
refs/tags/v0.10.1
```

vkgnosis: I'll test it in a fork

nlordell: A simple solution in that case would be to:
```
$ echo $GITHUB_REF
refs/tags/v0.10.1
$ echo ${GITHUB_REF#refs/tags/}
v0.10.1
```

**edit**: improved variable substitution

nlordell: This support page also seems to suggest that its `ref/tags/$TAGNAME`: https://github.community/t/how-to-get-just-the-tag-name/16241

nlordell (approved):
This is awesome! Thanks for doing all the Travis -> Github Actions migration work! Once you've verified the `GITHUB_REF` value feel free to:
- Remove the `--dry-run`, YOLO!
- Remove the `echo $GITHUB_REF`

Just make sure to run `ci/bump.sh v0.10.1` (I don't mind if its a part of this PR as well) before creating the release since the deploy script checks that the Cargo version matching the tag.

vkgnosis:
How about this @nlordell ? I have removed the variable defaults in the script so that we always pass via arguments. Either way github doesn't leak the secret because its redacted in the output. I thought that was a bit cleaner because then the script doesnt need CI specific instructions.

nlordell:
I really like the suggestion! Maybe just add a:
```bash
if  [[ -z "$tag" || -z "$token" ]]; then
	>&2 echo "ERROR: missing tag and/or token parameters"
	exit 1
fi
```

nlordell:
@vkgnosis Feel free to bump to v0.10.1 in this PR so we can try the deployment right away.

nlordell:
Nightly clippy error :cry: :

```
 error: internal compiler error: compiler/rustc_metadata/src/rmeta/decoder.rs:1189:17: get_optimized_mir: missing MIR for `DefId(62:268 ~ web3[1e8d]::api::eth_filter::{impl#9}::create_blocks_filter::{closure#0})`

```

nlordell:
NVM: 
> GitHub App like Mergify are not allowed to merge pull request where .github/workflows is changed.

PR #462, opened 2021-02-03
Allow deployment override to specify transaction hash

vkgnosis:
The transaction is part of the generated artifact and for example used
to lower bound the first block on which an event can have occurred.


ethcontract-generate/Cargo.toml
```diff
@@ -20,3 +20,4 @@ proc-macro2 = "1.0"
 quote = "1.0"
 syn = "1.0.12"
 url = "2.1"
+web3 = { version = "0.15", default-features = false }
```
nlordell: I think this should be re-exported from `ethcontracts-common` and then reused from there (kind of how its done for `Address`). My main motivation for that is that it:
- means there are less instances of `web3` versions that need to be updated
- it in principle decouples the `web3` dependency from `ethcontract-generate` (you can imagine that `H256` is swapped for an alternate implementation).

nlordell (changes requested):
Looks great! Just one minor change request around how the `H256` type is made available to the generating crate.

ethcontract-generate/Cargo.toml
vkgnosis: Agree. Changed it to rexport TransactionHash.

vkgnosis:
Is it okay if I bump the version in this PR too? Technically this a breaking change because the signature of the function has changed. Should I bump minor or patch version?

nlordell (approved):

ethcontract-generate/src/contract/common.rs
```diff
@@ -174,6 +175,20 @@ fn expand_address(address: Address) -> TokenStream {
...
+/// with quasi-quoting for code generation.
+fn expand_transaction_hash(hash: Option<TransactionHash>) -> TokenStream {
+    let hash = match hash {
+        Some(hash) => hash,
+        None => return quote! { None },
+    };
+
+    let bytes = hash.as_bytes().iter().copied().map(Literal::u8_unsuffixed);
+    quote! {
+        self::ethcontract::TransactionHash([#( #bytes ),*])
```
nlordell, line 188: This might require re-exporting `H256 as TransactionHash` in `ethcontract/src/lib.rs` as well.

nlordell:
> Should I bump minor or patch version?

Since it breaks backwards compatibility I would say to do a minor bump, but no strong opinion either way.

PR #463, opened 2021-02-03
Fix missing Some in deployment transaction hash

vkgnosis:
Serves me right for not making sure that an example uses this.

### DIFF
--- a/ethcontract-common/Cargo.toml
+++ b/ethcontract-common/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "ethcontract-common"
-version = "0.11.0"
+version = "0.11.1"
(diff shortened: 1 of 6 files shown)


fleupold (approved):
Does this mean 0.11.0 broken (or only if one tries to use the features)? Maybe we can then also include #464 in the next release to not cut too many minor versions (although it doesn't really matter).

vkgnosis:
Only broken if the feature was used.

vkgnosis:
We can release both together as 0.12.0. When your PR is ready, you or I can merge both, and bump the version.

nlordell (approved):

nlordell:
I suggest bumping to `0.11.1`, but not a strong opinion either way.

PR #477, opened 2021-03-11
Fix revert reason handling for geth nodes

### LINKED ISSUE #466: Decode geth revert messages
See https://github.com/gnosis/dex-services/issues/1647 . Want to track this here from now.

vkgnosis:
Fixes #466


nlordell (approved):
Very nice!

PR #497, opened 2021-04-15
Implement Tokenize for BytesArray & update generator

ducthotran2010:
The contract call could return `uint8[]` instead of `bytes`, but `uint8[]` is generated as `Vec<u8>` and will always be used as `Token::Bytes` and will always return `Error::TypeMismatch`: https://github.com/gnosis/ethcontract-rs/blob/main/ethcontract/src/tokens.rs#L63-L77

The `uint8[]` in this PR will be generated as `web3::BytesArray` with `Tokenize` implemented.

### DIFF
--- a/ethcontract-common/src/lib.rs
+++ b/ethcontract-common/src/lib.rs
@@ -14,8 +14,8 @@ pub use crate::bytecode::Bytecode;
 pub use crate::truffle::Artifact;
 pub use ethabi_fork_ethcontract::{self as abi, Contract as Abi};
 use serde::Deserialize;
-pub use web3::types::Address;
 pub use web3::types::H256 as TransactionHash;
+pub use web3::types::{Address, BytesArray};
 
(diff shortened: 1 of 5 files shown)


vkgnosis:
Thanks for the PR.
For context, this is also mentioned in the comment at https://github.com/gnosis/ethcontract-rs/blob/be7205104ed2f0a99be07ae7edafa3ec066a8cdc/ethcontract/src/tokens.rs#L1 . Still thinking about the best way to handle it.
I'd also like to add this to one of the examples as a test case (probably the abi one). No need for you to do it in this PR (but you can if you want to).

nlordell:
I believe this has been fixed with https://github.com/gnosis/ethcontract-rs/pull/498

vkgnosis:
Yes

nlordell:
@ducthotran2010 feel free to re-open this if you have any issues with the fix in #498 (should be included in the latest 0.12 release).

PR #498, opened 2021-04-20
Allow arrays of u8 by moving bytes to a wrapper type

vkgnosis:
By wrapping Bytes and FixedBytes we make the code and api more consistent because Vec<u8> is no longer a special case. Fixes not being able to use Array of uint8 in abi.


nlordell (approved):
The change looks very good so far.

vkgnosis:
Fixed the event issue. Ready to merge now.

PR #521, opened 2021-05-31
Evict cache in attempt to fix CI

vkgnosis:
For some reason CI succeeds on PRs but not on main. I suspect the reason is cache related. So with this PR I am changing the compiler version which should evict the cache.
Also updates the minimum supported rust version in the readme which was behind. It's not great to change the minimum version just to fix CI but we have also always bumped it when needed so should be fine. And we could go down a version again after the cache got cleared if we wanted to.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 [![Build Status](https://travis-ci.org/gnosis/ethcontract-rs.svg?branch=main)](https://travis-ci.org/gnosis/ethcontract-rs)
 [![Crates.io](https://img.shields.io/crates/v/ethcontract.svg)](https://crates.io/crates/ethcontract)
 [![Docs.rs](https://docs.rs/ethcontract/badge.svg)](https://docs.rs/ethcontract)
-[![Rustc Version](https://img.shields.io/badge/rustc-1.47+-lightgray.svg)](https://blog.rust-lang.org/2019/12/19/Rust-1.47.0.html)
+[![Rustc Version](https://img.shields.io/badge/rustc-1.52+-lightgray.svg)](https://blog.rust-lang.org/2021/05/06/Rust-1.52.0.html)
 
 # `ethcontract-rs`
 
@@ -30,7 +30,7 @@ for interacting with contract functions in a type-safe way.
 
 ### Minimum Supported Rust Version
 
-The minimum supported Rust version is 1.42.
+The minimum supported Rust version is 1.52.
 
 ## Generator API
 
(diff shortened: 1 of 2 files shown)


nlordell:
Instead of changing the Rust version, we could just add a `key` to force cache eviction:
https://github.com/Swatinem/rust-cache#inputs

nlordell:
Also, looks like Rust beta is failing on some clippy warnings:
```
error: methods with the following characteristics: (`to_*` and `self` type is `Copy`) usually take `self` by value
   --> ethcontract/src/int.rs:502:26
    |
502 |     pub fn to_big_endian(&self, bytes: &mut [u8]) {
    |                          ^^^^^
    |
    = note: `-D clippy::wrong-self-convention` implied by `-D warnings`
    = help: consider choosing a less ambiguous name
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#wrong_self_convention

error: methods with the following characteristics: (`to_*` and `self` type is `Copy`) usually take `self` by value
   --> ethcontract/src/int.rs:507:29
    |
507 |     pub fn to_little_endian(&self, bytes: &mut [u8]) {
    |                             ^^^^^
    |
    = help: consider choosing a less ambiguous name
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#wrong_self_convention
```

I think in this case, its better to just ignore it so that the `I256` type has the same interface as `U256`.

vkgnosis:
The beta and nightly builds shouldn't matter for CI failing. The real failure is in the 1.51 build on main in cargo test when linking the code.

vkgnosis:
Hmm so now the build even fails here. Maybe it is a real issue with that rust version at least on github. Let me try 1.52 again.
https://github.com/gnosis/ethcontract-rs/pull/521/checks?check_run_id=2709525019

nlordell:
> The beta and nightly builds shouldn't matter for CI failing

Just meant something to fix while you're at it cleaning up the CI :smile:

vkgnosis:
Mostly I want it to show green on main. Will maybe look into beta and nightly in another PR. Then make a new release.

nlordell (approved):

README.md
```diff
@@ -1,7 +1,7 @@
 [![Build Status](https://travis-ci.org/gnosis/ethcontract-rs.svg?branch=main)](https://travis-ci.org/gnosis/ethcontract-rs)
 [![Crates.io](https://img.shields.io/crates/v/ethcontract.svg)](https://crates.io/crates/ethcontract)
 [![Docs.rs](https://docs.rs/ethcontract/badge.svg)](https://docs.rs/ethcontract)
-[![Rustc Version](https://img.shields.io/badge/rustc-1.47+-lightgray.svg)](https://blog.rust-lang.org/2019/12/19/Rust-1.47.0.html)
+[![Rustc Version](https://img.shields.io/badge/rustc-1.52+-lightgray.svg)](https://blog.rust-lang.org/2021/05/06/Rust-1.52.0.html)
```
nlordell:
```suggestion
[![Rustc Version](https://img.shields.io/badge/rustc-1.52+-lightgray.svg)](https://blog.rust-lang.org/2021/05/06/Rust-1.52.1.html)
```

PR #523, opened 2021-05-31
Make Bytes wrapper Serialize

vkgnosis:
We allow adding custom derives like Serialize to generated Event
structs but Bytes wasn't Serialize itself.

### DIFF
--- a/ethcontract/src/tokens.rs
+++ b/ethcontract/src/tokens.rs
@@ -15,6 +15,7 @@
 use crate::I256;
 use arrayvec::ArrayVec;
 use ethcontract_common::{abi::Token, TransactionHash};
+use serde::{Deserialize, Serialize};
 use std::convert::TryInto;
 use web3::types::{Address, U256};
 
@@ -51,7 +52,9 @@ pub trait Tokenize {
 
 /// Wrapper around Vec<u8> and [u8; N] representing Token::{Bytes, FixedBytes}. Distinguishes a list
 /// of u8 from bytes.
-#[derive(Clone, Copy, Debug, Default, Eq, Hash, PartialEq, PartialOrd, Ord)]
+#[derive(
+    Clone, Copy, Debug, Default, Deserialize, Eq, Hash, PartialEq, PartialOrd, Ord, Serialize,
+)]
 pub struct Bytes<T>(pub T);
 
 impl Tokenize for Bytes<Vec<u8>> {


fleupold (approved):

ethcontract/src/tokens.rs
```diff
@@ -51,7 +52,9 @@ pub trait Tokenize {
 
 /// Wrapper around Vec<u8> and [u8; N] representing Token::{Bytes, FixedBytes}. Distinguishes a list
 /// of u8 from bytes.
-#[derive(Clone, Copy, Debug, Default, Eq, Hash, PartialEq, PartialOrd, Ord)]
+#[derive(
+    Clone, Copy, Debug, Default, Deserialize, Eq, Hash, PartialEq, PartialOrd, Ord, Serialize,
+)]
```
nlordell, line 57: Does this need [`serde(transparent)`](https://serde.rs/container-attrs.html#transparent)? IOr, does this need to do hex serialization/deserialization (for compatibility with Ethereum JSON RPC)?

vkgnosis: I don't think this "needs" anything because it isn't serde::Seralized by ethcontract or web3. This is because for users of ethcontract it is convenient to be able to derive Serlialize on event structs like we did in v1 to store them on disk.

vkgnosis: In my update web3 PR to services I had to remove the Serialize derive from the build script because of this (but also we don't need it to be serializable atm).

nlordell: Good point. I thought it needed to deserialize some `web3` related JSON. My only concern is that it conflicts a bit with the [`web3::types::Bytes` serialization implementation](https://docs.rs/web3/0.16.0/src/web3/types/bytes.rs.html#17-64) which might surprise users.

vkgnosis: Alternatively we could just have it not be serializable and if users want to serialize an event struct they'll have to create their own wrapper. For example if we didn't already depend on serde in this crate I wouldn't have liked to add the dependency just to make this Serializable.

nlordell: We do because `web3` and `jsonrpc-core` both depend on `serde`.

vkgnosis: So what do you think? Serializable or not?

nlordell (approved):
Lets just leave it `Serialize` for now, as you suggested (so it can be more generically used with other formats). Users can do `web3::types::Bytes(value.0)` if they want [Ethereum JSON RPC compliant `DATA` hex encoding](https://eth.wiki/json-rpc/API#hex-value-encoding).

We can always change this to use the `web3::types::Bytes` implementation if this becomes a pain point.

Issue #534, opened 2021-06-14
Add exponential backoff when polling for transaction confirmation


taminomara:
Right now, we poll block number once every 7 seconds. With fix for #533 that may become an issue. To speed things up, I propose we add exponential backoff, starting at `250ms` (or maybe less? maybe that should depend on the number of blocks we're waiting and the average speed of mining?) and going up to `7s`.

PR #535, opened 2021-06-14
Use polling instead of filters to wait for tx confirmation

### LINKED ISSUE #533: Ethcontract erroneously waits for transaction even after it was mined
From slack discussion:

@taminomara:
> I've noticed a couple of times when executing methods via ethcontract that they may hang and never resolve even though transaction gets mined successfully (I'm using local chain running in dev mode).

@nlordell:
> So, what happens is that:
> - Transaction is sent by ethcontract
> - Block number is retrieved and is block X
> - transaction gets mined and is now block X+1
> - Transaction receipt gets retrieved with block number X+1
> - ethcontract thinks that it needs to wait for 1 block
> - creates a eth_newBlockFilter to get notifications about new blocks
> - waits for ever to get notified about new blocks, but they don't come because Geth only mines when a new transcation comes is.

@taminomara:
> Interesting. I wonder why we didn't catch it sooner? I guess geth mines some additional blocks?

@nlordell:
> I think its because:
> - On Ganache, the transaction gets mined immediately, so the race is not possible
> - On real nodes, new blocks eventually come in, so worst case we wait an extra block

taminomara:
There's a race condition. If transaction gets mined after we fetch latest block number, but before we create a filter, then we wait for more blocks than necessary. If new blocks are not generated (i.e. dev setup or testchains with low activity), waiting for transaction may hang.

We attempt to remedy this issue by switching entirely to polling for block numbers. In theory, this should also decrease node's load. A potential downside is that confirmation times may increase. That is, we now poll node every seven seconds instead of confirming transaction immediately after the target block was mined. We will fix that later by introducing exponential backoff to polling mechanism — see #534.

Fix #533.

### DIFF
--- a/ethcontract/src/transaction/confirm.rs
+++ b/ethcontract/src/transaction/confirm.rs
@@ -233,13 +233,7 @@ mod tests {
         // transaction pending
         transport.add_response(json!("0x1"));
         transport.add_response(json!(null));
-        // filter created
-        transport.add_response(json!("0xf0"));
-        // polled block filter for 1 new block
-        transport.add_response(json!([]));
-        transport.add_response(json!([]));
-        transport.add_response(json!([H256::repeat_byte(2)]));
-        // check transaction was mined
+        // poll for one block
         transport.add_response(json!("0x2"));
         transport.add_response(generate_tx_receipt(hash, 2));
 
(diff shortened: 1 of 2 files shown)


ethcontract/src/transaction/confirm.rs
```diff
@@ -125,27 +120,31 @@ impl<T: Transport> ConfirmationContext<'_, T> {
...
                 Some((tx_block, tx)) => {
                     let target_block = tx_block + self.params.confirmations;
-                    let remaining_confirmations = target_block.saturating_sub(latest_block);
 
-                    if remaining_confirmations.is_zero() {
+                    // This happens in two cases:
+                    // - we don't need additional confirmation, transaction receipt is enough,
+                    // - the transaction was mined before we queried `latest_block`, thus
+                    //   `latest_block >= tx_block`.
+                    if latest_block >= target_block || self.params.confirmations == 0 {
```
nlordell: Nice! This will solve the race we saw.

nlordell (commented):
Looks good so far! I like how it cut down on complexity.

vkgnosis (approved):
Looks good to me but let's wait for Nick's approval before merging.

taminomara:
For context, openethereum team confirmed that polling generates less load for node than filters ([slack thread]).

[slack thread]: https://gnosisinc.slack.com/archives/C0107GMHWKY/p1623675227061700

vkgnosis:
Also in practice at least with our load balanced set ups filters not great because every request might go to a new node which doesn't have the filter set up yet.

nlordell (approved):

PR #538, opened 2021-06-15
Add exponential backoff for transaction confirmation polling

### LINKED ISSUE #534: Add exponential backoff when polling for transaction confirmation

taminomara:
By switching to polling method to confirm transaction, we've introduced a potential issue with confirmation wait times.

On average, a new block appears every seven or so seconds, so we query for current block number with this frequency. This is fine when we need to wait for several blocks. However it we're only waiting for one or two blocks, seven seconds seems excessive. A new block may appear sooner, and we don't want to waste too much time on waiting.

To fix this, we introduce exponential backoff to the polling mechanism. We start with delay of `250ms`, and gradually increase it to `7s`.

Fix #534

### DIFF
--- a/ethcontract/src/transaction/confirm.rs
+++ b/ethcontract/src/transaction/confirm.rs
@@ -9,6 +9,7 @@
 use crate::errors::ExecutionError;
 use crate::transaction::TransactionResult;
 use futures_timer::Delay;
+use std::cmp::min;
 use std::time::Duration;
 use web3::api::Web3;
 use web3::types::{TransactionReceipt, H256, U64};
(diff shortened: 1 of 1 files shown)


ethcontract/src/transaction/confirm.rs
```diff
@@ -71,12 +88,30 @@ impl ConfirmParams {
         self
     }
 
-    /// Set new value for [`poll_interval`].
+    /// Set new value for [`poll_interval_min`].
     ///
-    /// [`poll_interval`]: #structfield.poll_interval
+    /// [`poll_interval_min`]: #structfield.poll_interval_min
     #[inline]
-    pub fn poll_interval(mut self, poll_interval: Duration) -> Self {
```
nlordell, line 78: Do you think there is value in adding a :
```rust
     pub fn poll_interval(mut self, min: Duration, max: Duration, factor: f32) -> Self;
```

method as well?

ethcontract/src/transaction/confirm.rs
```diff
@@ -24,23 +25,37 @@ pub struct ConfirmParams {
...
 #[cfg(not(test))]
-pub const DEFAULT_POLL_INTERVAL: Duration = Duration::from_secs(7);
+const DEFAULT_POLL_INTERVAL_MAX: Duration = Duration::from_millis(7000);
 #[cfg(test)]
-pub const DEFAULT_POLL_INTERVAL: Duration = Duration::from_secs(0);
+const DEFAULT_POLL_INTERVAL_MAX: Duration = Duration::from_millis(0);
+
+/// Default factor for increasing delays between node polls.
+#[cfg(not(test))]
+const DEFAULT_POLL_INTERVAL_FACTOR: f32 = 1.7;
```
nlordell, line 56: Just curious about the choice of `f32` over `f64`, not that it really matters :stuck_out_tongue:.

ethcontract/src/transaction/confirm.rs
```diff
@@ -189,13 +224,20 @@ impl<T: Transport> ConfirmationContext<'_, T> {
...
-            delay(self.params.poll_interval).await;
-            let latest_block = self.web3.eth().block_number().await?;
+            delay(cur_delay).await;
 
+            let latest_block = self.web3.eth().block_number().await?;
             if target_block <= latest_block {
                 break Ok(latest_block);
             }
+
+            cur_delay = min(
```
nlordell, line 246: Should we reset the delay on every new block? My concern is that if we set a high number of confirmations (like 20 blocks for example) then we will always be polling at `poll_interval_max`.

nlordell (commented):
Nice change.

Could you add a description to the PR before merging? We typically use this template for PRs:
> [Optional] 
> Fixes #issue
>
> *<Context of what this change does, why it is needed and how it is accomplished>*
>
> ### Test Plan
>
> *<Explain how you and a reviewer have/intend to test this change>*

The main motivation here is that the description gets included in the commit message when squash merging so that we get the details directly in Git (instead of just having an extra level of indirection as a reference to an Github issue that needs to be looked up).

(we should probably add the PR template as we did in the other repos).

ethcontract/src/transaction/confirm.rs
taminomara: Idk, we don't need any precision here, so I choose the smallest thing possible. Sort of like the default int is `i32`.

ethcontract/src/transaction/confirm.rs
taminomara: Given that the average block time is 14 seconds, I think polling once every 7 seconds is enough. The smaller delays will only factor when we wait for one, maybe two blocks.

taminomara: That makes sense. I'd reset it to the average delay between blocks instead of min delay.

P.S. My understanding of block times was wrong, disregard the deleted comment.

taminomara: Wait no, I've confused myself. Ok, my point is, if new blocks appear slower than `poll_interval_max`, then there's no need to reset the delay.

taminomara: Since `cur_delay` is limited by `max_delay`, I don't think we should reset it. A new block appears every five to fifteen seconds, so wait time of seven seconds is good.

taminomara:
I've added a description.

ethcontract/src/transaction/confirm.rs
nlordell: Cool just curious if there was a specific reason for it.

ethcontract/src/transaction/confirm.rs
nlordell: Sounds good to me. I think its a valid point that its bound by a max polling interval, and with large block confirmations you will be waiting a long time anyway.

nlordell (approved):
Great! And thanks for the nice PR description.

vkgnosis (approved):

Issue #512, opened 2021-05-17
Add support for `hardhat` and `hardhat-deploy` artifacts and deployments.


nlordell:
Since we are migrating more and more of our repositories to `hardhat` and `hardhat-deploy`, it would be nice to support this natively.

- [x] Rename `Artifact` to `Contract`.
- [x] Introduce `struct Artifact`.
- [X] Implement loader for truffle format.
- [x] Implement loader for hardhat export format.
- [x] #196
- [x] #170
- [x] Rework `Builder` to work with `Contract`, implement `ContractBuilder`.
- [x] Support artifact formats in macro.
- [x] Merge and release `ethabi`.
- [x] Merge feature branch into `main`.

taminomara:
I want to look into this during Ecosystem Fridays maybe?

taminomara:
I'll start with adding this to `ethcontract-generate`, then move on to macro.

taminomara:
So, hardhat artefacts are structured differently from truffle artefacts. In hardhat, you get a list of contracts for each network, while in truffle it's a list of networks for each contract. This difference has two major impacts. First, in hardhat it's possible that a contract will have different ABIs on different networks. Second, we have a single json file containing all info about all contracts. So, potentially, we can generate multiple contracts with a single macro invocation.
I don't want to support contracts that have different ABIs on different networks, I think it's ok to fail compilation when we see this.

As for generating multiple contracts from a single macro invocation, I think this could be useful; we won't have to load and parse the same file multiple times. However, that will require changing the interface of both builder and macro.

I'm thinking about what a good interface will look like.

nlordell:
> I don't want to support contracts that have different ABIs on different networks, I think it's ok to fail compilation when we see this.

I also agree. I think having multiple contracts with different ABIs is a huge can of worms that I don't want to open.

Also, as long as we can have a way to specify a single deployment, we can techincally still support different ABIs on different networks by having multiple `contract!` macro invocations.

> So, potentially, we can generate multiple contracts with a single macro invocation.

Just to make sure I understood correctly, would the suggestion be that we can specify multiple paths to JSON artifacts and then the network information would just be merged (erroring if non-mergable differences - such as different ABIs are encountered)?

taminomara:
> we can specify multiple paths to JSON artifacts and then the network information would just be merged

My line of thought was the following. One artifact (that is, one JSON file) may now export multiple contracts. Thus, we should be able to generate all of them in one go.

[...]

PR #544, opened 2021-06-22
Add Artifact trait

taminomara:
Part of #512.

We're adding the `Artifact` trait that will provide a uniform interface for working with different artifact sources.

### DIFF
--- a/ethcontract-common/src/artifact.rs
+++ b/ethcontract-common/src/artifact.rs
@@ -10,9 +10,9 @@
 
 use crate::errors::ArtifactError;
 use crate::Contract;
+use std::collections::HashMap;
 
 pub mod truffle;
-pub mod hardhat;
 
 /// An entity that contains compiled contracts.
 pub trait Artifact {
--- a/ethcontract-common/src/artifact.rs
+++ b/ethcontract-common/src/artifact.rs
@@ -8,93 +8,94 @@
 //! artifact models. It also provides tools to load artifacts from different
 //! sources, and parse them using different formats.
 
-use crate::errors::ArtifactError;
 use crate::Contract;
 use std::collections::HashMap;
 
 pub mod truffle;
 
 /// An entity that contains compiled contracts.
-pub trait Artifact {
-    /// Describes where this artifact comes from. This could be anything:
-    /// path to a json file, url, or something else. This function is used
-    /// in error messages.
-    fn origin(&self) -> &str;
-
-    /// Get contract by name. Pass an empty string to get an unnamed contract
-    /// if an artifact implementation supports it.
-    fn contract(&self, name: &str) -> Result<Option<&Contract>, ArtifactError>;
-
-    /// Iterate over contracts in the artifact. Order of iteration
-    /// is not specified.
-    fn iter(&self) -> Box<dyn Iterator<Item = &Contract> + '_>;
-}
-
-/// A simple [`Artifact`] implementation that only holds one contract.
-///
-/// This is used to represent artifacts similar to truffle and waffle.
-pub struct SimpleArtifact {
+pub struct Artifact {
     origin: String,
-    contract: Contract,
+    contracts: HashMap<String, Contract>,
 }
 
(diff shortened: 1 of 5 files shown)


ethcontract-common/src/artifact.rs
```diff
@@ -0,0 +1,97 @@
+//!
+
+use crate::errors::ArtifactError;
+use crate::Contract;
+
+/// An entity that contains compiled contracts.
+pub trait Artifact {
```
nlordell: I like this abstraction. If only there were HKT you could use an `Iter` associated type :stuck_out_tongue:.

ethcontract-common/src/artifact.rs
```diff
@@ -0,0 +1,114 @@
...
+            Ok(Some(&self.contract))
+        } else {
+            Ok(None)
+        }
+    }
+
+    fn iter<'a>(&'a self) -> Box<dyn Iterator<Item = &Contract> + 'a> {
+        Box::new(SimpleArtifactIter {
+            contract: Some(&self.contract),
+        })
```
nlordell: Counldn't this just use [`std::iter::once`](https://doc.rust-lang.org/std/iter/fn.once.html)?.

ethcontract-common/src/artifact/truffle.rs
```diff
@@ -0,0 +1,70 @@
...
+        Ok(SimpleArtifact::with_origin(origin.to_string(), contract))
+    }
+
+    /// Loads a truffle artifact from disk.
+    pub fn load_from_file(&self, path: &Path) -> Result<SimpleArtifact, ArtifactError> {
+        let origin = self
+            .origin
+            .as_deref()
+            .or_else(|| path.to_str())
+            .unwrap_or("<filesystem>");
```
nlordell: Can we use [`Path::display`](https://doc.rust-lang.org/std/path/struct.Path.html#method.display) which does a lossy conversion to string and might give more context than "<filesystem>" in case of non-Unicode data?

nlordell (approved):
Nice abstraction!

vkgnosis (approved):

ethcontract-common/src/artifact.rs
taminomara: The richness of standard library never ceases to amaze me!

taminomara: So `std::iter::once` creates `std::iter::Once` that wraps `core::option::IntoIter` that wraps `core::option::Item` that wraps `core::option::Option` 🤔

taminomara:
I wonder it we're actually overdoing on abstractions here. Maybe a struct that maps contract names into contracts is enough? This will not allow loading contracts from hardhat deploy directory lazily, but is it a feature that we really want? I know some projects have lots of contracts, but aren't they going to be loaded anyways?

vkgnosis:
I don't have much experience with the low level artifact stuff from hardhat etc so I'm not sure. Not supporting lazy loading seems fine but it depends on how expensive "loading" is. Maybe we should start with the simpler (less abstracted) version and re-evaluate later?

taminomara:
I've tried to implement lazy loader for hardhat and realised that it makes code way too complicated, and without any good reason. So I've re-implemented `Artifact` as a struct.

ethcontract-common/src/artifact/truffle.rs
```diff
@@ -0,0 +1,133 @@
...
+    pub fn load_contract_from_file(&self, path: &Path) -> Result<Contract, ArtifactError> {
+        let mut contract: Contract = serde_json::from_reader(File::open(path)?)?;
+        if let Some(name) = &self.name {
+            contract.name = name.clone();
+        }
+        Ok(contract)
+    }
+}
+
+impl Default for TruffleLoader {
```
nlordell, line 129: ubernit: I think this can be derived.

nlordell (approved):
Looks nice. Treating an artifact as data instead of behaviour also works well IMO.

PR #546, opened 2021-06-23
Remove compatibility with old ethabi parsing

taminomara:
Solc v0.6 introduced `receive` function type. With ethabi 14.0, this type is finally supported, so we can implement `receive` method and drop compatibility layer from our parser.

### DIFF
--- a/ethcontract-common/src/contract.rs
+++ b/ethcontract-common/src/contract.rs
@@ -17,7 +17,6 @@ pub struct Contract {
     #[serde(rename = "contractName")]
     pub contract_name: String,
     /// The contract ABI
-    #[serde(with = "compat_solc_v0_6")]
     pub abi: Abi,
     /// The contract deployment bytecode.
     pub bytecode: Bytecode,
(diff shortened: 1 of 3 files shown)


taminomara:
Note: there's [a report] that the issue is not fixed, but I wasn't able to confirm it.

[a report]: https://github.com/rust-ethereum/ethabi/issues/185#issuecomment-846460019

taminomara:
Solc changelog says:

> The unnamed function commonly referred to as “fallback function” was split up into a new fallback function that is defined using the fallback keyword and a receive ether function defined using the receive keyword.
> - If present, the receive ether function is called whenever the call data is empty (whether or not ether is received). This function is implicitly payable.
> - The new fallback function is called when no other function matches (if the receive ether function does not exist then this includes calls with empty call data). You can make this function payable or not. If it is not payable then transactions not matching any other function which send value will revert. You should only need to implement the new fallback function if you are following an upgrade or proxy pattern.

From what I understand, on our side we're just sending a call with empty call data, the node will figure out which method should be called. Is this correct? If so, do we maybe want to rename this method in our API? Or replace it with `receive` method that doesn't accept any arguments? Or create a separate `receive` method whilst leaving `fallback` method as it is?

nlordell:
>  Is this correct?

`fallback` and `receive` functions both accept calldata. The SC bytecode looks at the first 4 bytes of the calldata and uses it to determine which function to jump to when its executing. If the 4 bytes are either not there or don't match any known function, then the SC will jump to the `fallback` function if the value (the amount of Ether being transferred) is `0` and to the `receive` function otherwise.

> If so, do we maybe want to rename this method in our API?

Hmm, the issue is that to actually call the `receive` function you need to specify a non-`0` value, otherwise it will call the `fallback` function. Since we don't differentiate between `payable` and non-`payable` functions, I think its fine to conflate `receive` and `fallback` together. If we really want to support `receive`, then we would need a separate `receive` function that takes a non-`0` value input, otherwise it would just never get called.

nlordell (approved):

nlordell:
Super happy to get rid of that hacky code!

taminomara:
> If the 4 bytes are either not there or don't match any known function, then the SC will jump to the fallback function if the value (the amount of Ether being transferred) is 0 and to the receive function otherwise.

Hmm, solc doc specifically mentions this: If present, the receive ether function is called whenever the call data is empty (whether or not ether is received).

taminomara:
I wonder if we really need `fallback` function. It feels like `receive` will cover all use cases.

nlordell:
Ah I see, `receive` is called on empty calldata, while `fallback` accepts arbitrary calldata when the selector (first 4 bytes) doesn't match one of the expected functions.

vkgnosis (approved):

taminomara:
So, do we want to rename anything, or do I merge this PR?

nlordell:
I think its fine for now to conflate `fallback` with `receive` for now.

PR #548, opened 2021-06-24
Implement artifact loader for hardhat export

taminomara:
Part of #512.

Implements artifact loader for hardhat export (`hardhat export` command).


ethcontract-common/src/artifact/hardhat.rs
```diff
@@ -0,0 +1,290 @@
...
+
+    contracts: HashMap<String, ContractWithAddress>,
+}
+
+#[derive(Clone, Debug, Deserialize)]
+struct ContractWithAddress {
+    address: Address,
+    #[serde(flatten)]
+    contract: Contract,
+}
```
nlordell, line 325: Could you add some unit tests that maybe `load_from_json` with some `json!` value? It would be nice as it would also document the format a little in code.

ethcontract-common/src/artifact.rs
```diff
@@ -99,3 +124,51 @@ impl Default for Artifact {
...
+    /// Reference to the newly inserted contract.
+    pub inserted_contract: ContractMut<'a>,
+
+    /// If insert operation replaced an old contract, it will appear here.
+    pub old_contract: Option<Contract>,
+}
+
+/// A wrapper that allows mutating contract
+/// but doesn't allow changing its name.
+pub struct ContractMut<'a>(&'a mut Contract);
```
nlordell, line 149: This is needed because modifying the contract name would cause the `Artifact`'s `HashMap<String, Contract>` to become inconsistent right?

ethcontract-common/src/artifact.rs
```diff
@@ -99,3 +124,51 @@ impl Default for Artifact {
...
+        &mut self.0.bytecode
+    }
+
+    /// Get mutable access to networks.
+    pub fn networks_mut(&mut self) -> &mut HashMap<String, Network> {
+        &mut self.0.networks
+    }
+
+    /// Get mutable access to devdoc.
+    pub fn devdoc_mut(&mut self) -> &mut Documentation {
```
nlordell, line 168: One thing I was thinking - since the documentation in the contract is only really needed by the code generation, do you think it would be possible to separate it from `Contract` (so that it only contains things needed by the runtime) and fit it into the `Artifact` somehow.

Just rambling a little - haven't really thought this suggestion through.

nlordell (commented):
Code looks good.

Would just like to see a test that parses an `Artifact` from the Hardhat exported single artifact and multi artifact format. I think it would help catch regressions if the `Artifact` and `Contract` types change requiring the loader code to be adapted as well document the "shape" of the expected hardhat artifact in code.

ethcontract-common/src/artifact/hardhat.rs
```diff
@@ -0,0 +1,290 @@
...
+
+    /// Loads a hardhat artifact from disk.
+    pub fn load_from_file(&self, format: Format, path: &Path) -> Result<Artifact, ArtifactError> {
+        let origin = self
+            .origin
+            .clone()
+            .unwrap_or_else(|| path.display().to_string());
+        let mut artifact = Artifact::with_origin(origin);
+
+        let file = File::open(path)?;
```
vkgnosis: BufReader is usually faster and recommended by serde_json https://docs.serde.rs/serde_json/fn.from_reader.html

```suggestion
        let file = File::open(path)?;
        let reader = std::io::BufReader::new(file);
```

ethcontract-common/src/artifact/hardhat.rs
```diff
@@ -0,0 +1,290 @@
...
+    ///
+    /// [`denied networks`]: #structfield.networks_deny_list
+    #[inline]
+    pub fn deny_network(mut self, network: impl Into<String>) -> Self {
+        self.networks_deny_list.push(network.into());
+        self
+    }
+
+    /// Parse a hardhat artifact from JSON string.
+    pub fn load_from_string(&self, format: Format, json: &str) -> Result<Artifact, ArtifactError> {
```
vkgnosis: I am wondering which of these load_from makes sense. serde_json has https://docs.serde.rs/serde_json/index.html:

- from_reader
- from_slice
- from_str
- from_value

which is slightly different from what we have. Looking at the implementations of those functions it uses https://docs.serde.rs/serde_json/de/trait.Read.html to avoid code duplication. Maybe that would be useful here too.

This is not very important and don't want to sidetrack you too much, just wanted to mention.

vkgnosis (approved):

ethcontract-common/src/artifact/hardhat.rs
```diff
@@ -0,0 +1,290 @@
...
+            origin: Some(origin.into()),
+            networks_deny_list: Vec::new(),
+            networks_allow_list: Vec::new(),
+        }
+    }
+
+    /// Set new override for artifact's origin. See [`origin`] for more info.
+    ///
+    /// [`origin`]: #structfield.origin
+    #[inline]
```
vkgnosis: I don't strongly mind this but I feel that specifying inline is unnecessary. I feel it is better to leave it up to the compiler unless it is performance critical which I don't imagine for this struct.

taminomara: Rust still doesn't support inlining non-`#[inline]` functions across crates. That's fine if users of our library enable lto, but we can't guarantee this.

taminomara: Our strategy for using inline needs a discussion, but I'm ok with removing it here.

vkgnosis: What I mean is that I doubt inlining this matters compared to doing the much more expensive task of deserializing json afterwards. From my side no need to remove but wanted to mention it.

nlordell: Personally I'm OK either way. I think inlining makes perfect sense for these functions (and would be surprised if the LTO optimizer didn't do it even without the attribute).

That being said I also see @vkgnosis's point that it isn't strictly necessary as artifact parsing is probably not going to be in a performance critical section of code.

My personal opinion is that since its there and I believe its correct, we should leave it. However, I don't think in general we need to be strict with our use of `#[inline]` unless we are doing some performance critical code where inlining some function that is used in a tight loop makes a significant impact (and I would hope we would have benchmakrs in those cases to prove its usefulness).

taminomara: Since there are no other inline hints in the project, I'll remove them just to keep code style uniform. I agree that they don't do much difference, and it's compiler task to optimise these things. I agree that we shouldn't be strict about them.

ethcontract-common/src/artifact.rs
taminomara: We could wrap `Contract` into `ContractWithDocs`, but do we really need to? I mean, if we don't need those fields, we can just keep them empty? Besides, what if someone actually needs them, i.e. for UI purposes?

nlordell: I think you're right, lets leave it as is then.

taminomara:
Implemented review suggestions, working on tests.

ethcontract-common/src/artifact.rs
taminomara: Exactly.

taminomara:
Added tests, ready for next round of review.

ethcontract-common/src/artifact/hardhat.rs
```diff
@@ -0,0 +1,670 @@
...
+    }
+
+    /// Loads an artifact from disk.
+    pub fn load_from_file(&self, p: &Path) -> Result<Artifact, ArtifactError> {
+        let file = File::open(p)?;
+        let reader = BufReader::new(file);
+        self.load_artifact(p.display(), reader, from_reader, from_reader)
+    }
+
+    fn load_artifact<T>(
```
nlordell, line 175: Nice!

nlordell (approved):
Great! Thanks for the comprehensive tests :smile:.

vkgnosis (approved):

PR #558, opened 2021-07-01
Make signed transaction's hash available

vkgnosis:
Motivation is that to implement the archerdao integration I need access to the transaction hash. I would like to continue using ethcontract to do signing and this data is already available.

### Test Plan
adapted tests

### DIFF
--- a/ethcontract/Cargo.toml
+++ b/ethcontract/Cargo.toml
@@ -48,4 +48,5 @@ web3 = { version = "0.16", default-features = false, features = ["signing"] }
 zeroize = "1.1"
 
 [dev-dependencies]
+hex-literal = "0.3"
 tokio = { version = "1.6", features = ["macros"] }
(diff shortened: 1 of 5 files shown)


ethcontract/Cargo.toml
```diff
@@ -36,6 +36,7 @@ ethcontract-derive = { version = "0.12.2", path = "../ethcontract-derive", optio
 futures = "0.3"
 futures-timer = "3.0"
 hex = "0.4"
+hex-literal = "0.3"
```
nlordell: Should this be a dev-dependency?

nlordell (approved):

ethcontract/Cargo.toml
vkgnosis: Yes, will fix.

taminomara (approved):

PR #566, opened 2021-07-06
Fix CI

vkgnosis:
Another attempt to fix CI that worked locally. Through trial and error I found that running in these directories separately works in my local container setup.


taminomara:
Very weird that it works... I'll ask around rust discord server if anyone else noticed something similar.

taminomara (approved):

nlordell (approved):
Wow! Great find.

Issue #563, opened 2021-07-05
Implement loader for hardhat `deployments` directory format


taminomara:
- [x] Implement a loader.
- [ ] Support it in macro.

Macro support could be challenging. Right now `Source` assumes that an artifact could be represented as a single file, which is not the case here.

PR #567, opened 2021-07-07
Support loading artifact from hardhat deployments directory

taminomara:
Part of #563.

Allow loading artifacts from hardhat deployments directory by scanning its contents and parsing all allowed contracts.

### DIFF
--- a/ethcontract-common/src/artifact/hardhat.rs
+++ b/ethcontract-common/src/artifact/hardhat.rs
@@ -11,7 +11,15 @@
 //! all contracts deployed on it. It can be generated with
 //! `hardhat export --export-all` command.
 //!
-//! Both formats are supported by [`HardHatLoader`], see its documentation
+//! Third is hardhat's `deployments` directory. It contains more details about
+//! contracts than the previous two formats. Specifically, it has info about
+//! deployed bytecode, deployment transaction receipt, documentation for
+//! contract methods, and some other things. Given that it is a directory,
+//! there are obvious issues with loading it over network. For this reason,
+//! we don't recommend this export format for public libraries that export
+//! contracts.
+//!
+//! All three formats are supported by [`HardHatLoader`], see its documentation
 //! for info and limitations.
 //!
 //! [hardhat-deploy]: https://github.com/wighawag/hardhat-deploy
@@ -19,7 +27,7 @@
 use crate::artifact::Artifact;
 use crate::contract::Network;
 use crate::errors::ArtifactError;
-use crate::{Address, Contract};
+use crate::{Address, Contract, DeploymentInformation, TransactionHash};
 use serde::Deserialize;
 use serde_json::{from_reader, from_slice, from_str, from_value, Value};
 use std::collections::HashMap;
@@ -49,9 +57,6 @@ pub struct HardHatLoader {
     /// will be derived automatically.
     pub origin: Option<String>,
 
-    /// Artifact format.
-    pub format: Format,
-
     /// List of allowed network names and chain IDs.
     ///
     /// When loading a contract, networks with names that aren't found
@@ -76,20 +81,18 @@ pub struct HardHatLoader {
 
 impl HardHatLoader {
     /// Create a new hardhat loader.
-    pub fn new(format: Format) -> Self {
+    pub fn new() -> Self {
         HardHatLoader {
             origin: None,
-            format,
             networks_deny_list: Vec::new(),
             networks_allow_list: Vec::new(),
         }
     }
 
     /// Create a new hardhat loader and set an override for artifact's origins.
-    pub fn with_origin(format: Format, origin: impl Into<String>) -> Self {
+    pub fn with_origin(origin: impl Into<String>) -> Self {
         HardHatLoader {
             origin: Some(origin.into()),
-            format,
             networks_deny_list: Vec::new(),
             networks_allow_list: Vec::new(),
         }
@@ -100,12 +103,6 @@ impl HardHatLoader {
     /// [`origin`]: #structfield.origin
     pub fn origin(mut self, origin: impl Into<String>) -> Self {
         self.origin = Some(origin.into());
-        self
-    }
-
-    /// Set new format for artifacts.
-    pub fn format(mut self, format: Format) -> Self {
-        self.format = format;
         self
     }
 
@@ -127,7 +124,7 @@ impl HardHatLoader {
         self
     }
 
-    /// Add chain id to the list of [`denyid networks`].
+    /// Add chain id to the list of [`denied networks`].
     ///
     /// [`denied networks`]: #structfield.networks_deny_list
     pub fn deny_by_chain_id(mut self, network: impl Into<String>) -> Self {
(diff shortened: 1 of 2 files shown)


ethcontract-common/src/artifact/hardhat.rs
```diff
@@ -72,26 +77,44 @@ pub struct HardHatLoader {
...
+    /// Empty list means that all contracts are allowed.
+    pub contracts_allow_list: Vec<String>,
+
+    /// List of denied contract names.
+    ///
+    /// When loading artifact, loader will not load contracts if their names
+    /// are present in this list.
+    ///
+    /// Empty list means that no contracts are denied.
+    pub contracts_deny_list: Vec<String>,
```
bh2smith: Could you provide a reason or example use case for this?

ethcontract-common/src/artifact/hardhat.rs
```diff
@@ -130,7 +147,7 @@ impl HardHatLoader {
     /// Add chain id to the list of [`denyid networks`].
```
bh2smith, line 130: Denyied typo.

ethcontract-common/src/artifact/hardhat.rs
```diff
@@ -139,42 +156,158 @@ impl HardHatLoader {
...
             .push(NetworkEntry::ByName(network.into()));
         self
     }
 
-    /// Loads an artifact from a loaded JSON value.
-    pub fn load_from_reader(&self, v: impl Read) -> Result<Artifact, ArtifactError> {
-        self.load_artifact("<unknown>", v, from_reader, from_reader)
+    /// Add contract name to the list of [`allowed contracts`].
+    ///
+    /// [`allowed contracts`]: #structfield.contracts_allow_list
```
bh2smith: Not sure I am following this notation here on the left. 

Should I be reading this semantically "as the list of allowed contracts is available as a struct field at..."?

ethcontract-common/src/artifact/hardhat.rs
```diff
@@ -139,42 +156,158 @@ impl HardHatLoader {
...
         let path = p.as_ref();
         let file = File::open(path)?;
         let reader = BufReader::new(file);
-        self.load_artifact(path.display(), reader, from_reader, from_reader)
+        self.load_artifact(f, path.display(), reader, from_reader, from_reader)
+    }
+
+    /// Load an artifact from `deployments` directory.
+    pub fn load_from_directory(&self, p: impl AsRef<Path>) -> Result<Artifact, ArtifactError> {
+        self._load_from_directory(p.as_ref())
```
bh2smith, line 179: Is there some added benefit to adding a helper method with the same name for just this one line?

ethcontract-common/src/artifact/hardhat.rs
```diff
@@ -668,4 +848,109 @@ mod test {
...
+    #[test]
+    fn load_multi_deny_contract_name() {
+        let artifact = HardHatLoader::new()
+            .deny_contract("A")
+            .load_from_str(Format::MultiExport, MULTI_EXPORT)
+            .unwrap();
+
+        assert_eq!(artifact.len(), 1);
+
+        let a = artifact.get("B").unwrap();
```
bh2smith: Nit: given that we are explicitly denying contract "A" here. I would find this a lot less misleading if this other contract were called b. 

Looking at this more and along with my previous question. I don't even know why there are any artifacts at this point and certainly not something called "B" for any reason. Did you load the alphabet from string?

ethcontract-common/src/artifact/hardhat.rs
```diff
@@ -668,4 +848,109 @@ mod test {
...
+    }
+
+    #[test]
+    fn load_multi_deny_contract_name() {
+        let artifact = HardHatLoader::new()
+            .deny_contract("A")
+            .load_from_str(Format::MultiExport, MULTI_EXPORT)
+            .unwrap();
+
+        assert_eq!(artifact.len(), 1);
```
bh2smith: Perhaps I misunderstood the meaning of denied contracts but why is there artifact length now 1? All I see at this point is a new hardhat loader with a denied contract.

bh2smith (commented):
So far only reviewed hardhat dot rs. This is very long. Will have to come back later for more.

ethcontract-common/src/artifact/hardhat.rs
taminomara: It's just markdown hyperlink syntax:

```md
Text [with a link].

[with a link]: http://...
```

This syntax is extensively used across rust's `std` docs (see [`std/lib.rs`] for example).

[`std/lib.rs`]: https://github.com/rust-lang/rust/blob/fdd9a071474d898556891d5146fd9e8d1a10a077/library/std/src/lib.rs#L155-L189

ethcontract-common/src/artifact/hardhat.rs
taminomara: This is done to avoid code bloat. `impl AsRef<Path>` makes this method generic, so its body will be monomorphized multiple times depending on actual type of the argument. Also, the compiler will not emit assembly for generic functions when compiling them. Instead, it will save their IR and turn it to assembly whenever it sees the function used. This means that generics will get compiled multiple times, once per crate per argument type.

ethcontract-common/src/artifact/hardhat.rs
taminomara: We've created a loader with denied contract, then loaded an artifact. Artifact JSON contained two contract: `A` and `B`. Since we've denied loading `A`, the only contract that was loaded into the artifact is `B`, hence `artifact.len() == 1`.

taminomara:
I can move example into a separate PR, it that will help with review.

ethcontract-common/src/artifact/hardhat.rs
nlordell: This is from the `impl Trait` discussion here: https://gnosisinc.slack.com/archives/C01PBR5TWCS/p1624448800012900

ethcontract-common/src/artifact/hardhat.rs
nlordell: The issue, as I understand it, is that you can have contracts with the same name deployed with different bytecode across different networks with hardhat deployments (its not all or nothing like Truffle - which only keeps bytecode/ABI for the latest deployment). Those contracts need to be excluded because generating code for them is not possible (what if you have two functions with the same name but conflicting types in different ABIs for different versions of the same contract?).

ethcontract-common/src/artifact/hardhat.rs
```diff
@@ -139,42 +156,158 @@ impl HardHatLoader {
...
+                        )
+                    })?
+                    .to_string_lossy()
+                    .into_owned();
+
+                if !contract_name.ends_with(".json") {
+                    continue;
+                }
+
+                contract_name.truncate(contract_name.len() - ".json".len());
```
nlordell, line 278: Reusing the allocation :sunglasses:

ethcontract-common/src/artifact/hardhat.rs
vkgnosis: The reason is to make it more ergonomic for callers so that they don't have to use `as_ref()`. We had a discussion about this on slack in the rust channel a few weeks ago.

ethcontract-common/src/artifact/hardhat.rs
nlordell: It `.load_from_str(Format::MultiExport, MULTI_EXPORT)`, where `MULTI_EXPORT` is a Hardhat multi-contract export with two contracts `A` and `B`. Since `A` was denied, you would expect the loaded artifact to only have a single contract `B`.

ethcontract-common/src/artifact/hardhat.rs
nlordell: The loaded mulit-export is the JSON constant `MULTI_EXPORT`:

https://github.com/gnosis/ethcontract-rs/blob/e5722791cc6432884981ca672c0a1f2ef20bc7d8/ethcontract-common/src/artifact/hardhat.rs#L552-L580

ethcontract-common/src/artifact/hardhat.rs
```diff
@@ -668,4 +848,109 @@ mod test {
...
+    fn hardhat_dir() -> PathBuf {
+        let mut path = PathBuf::from(env!("CARGO_MANIFEST_DIR"));
+        path.push("../examples/hardhat/deployments");
+        path
+    }
+
+    #[test]
+    fn load_from_directory() {
+        dbg!(hardhat_dir());
+
```
nlordell: Not sure if leaving this in was intentional?

ethcontract-common/src/artifact/hardhat.rs
taminomara: Yeah, this was added in a previous PR. I didn't realise you can add a group of people as a reviewer, so I was just creating PRs without reviewers. You probably didn't get a notification, sorry about this. If you're interested, take a look at other PRs linked to #512, that will give more context about what's going on.

ethcontract-common/src/artifact/hardhat.rs
vkgnosis: I think the json contains a contract A and a contract B. In the previous test we use an allow list so only A gets loaded and in this test we use a deny list so only B gets loaded.

ethcontract-common/src/artifact/hardhat.rs
taminomara: Exactly. I'll add this to documentation.

vkgnosis (approved):

ethcontract-common/src/artifact/hardhat.rs
```diff
@@ -139,42 +156,158 @@ impl HardHatLoader {
...
+        self.load_artifact(f, path.display(), reader, from_reader, from_reader)
+    }
+
+    /// Load an artifact from `deployments` directory.
+    pub fn load_from_directory(&self, p: impl AsRef<Path>) -> Result<Artifact, ArtifactError> {
+        self._load_from_directory(p.as_ref())
+    }
+
+    fn _load_from_directory(&self, p: &Path) -> Result<Artifact, ArtifactError> {
+        let mut artifact = Artifact::with_origin(p.display().to_string());
```
nlordell, line 220: I'm super familiar with the hardhat directory layout, so this code was super easy to follow and understand. Maybe its worth including a comment blurb containing the "directory shape" for those who aren't so familiar. Alternatively, we could have a manually created test hardhat deployment directory that can be used with the unit test as well as document the expected "shape".

nlordell (approved):
The artifact loading code (`load_from_directory`) was very linear and easy to review.

One meta-PR comment, I think that the example could have been done in a separate PR, so this one could focus on the relevant `ethcontract-common/src/artifact/hardhat.rs` changes. I see that the unit test does depend on the new example output making it hard to split the two, but maybe its worth having manually created test hardhat deployment directory somewhere (kind of like an assertion of "this is the expected format and layout for a deployment directory").

nlordell:
> I can move example into a separate PR, it that will help with review.

TBH that would be kind of nice (if it isn't too much effort), I admittedly did not review the example carefully at all. If its too much work, don't worry about it.

taminomara:
Yeah this is difficult to review, I'll split the PR. I wish github allowed tree view the same way Intellij github plugin does.

taminomara:
I've moved example into a separate PR which I'll publish after this one. Filtering contracts by name and unrelated changes to documentation are moved to a separate PR as well.

taminomara:
From `+13,914 −10,346` to `+221 −99`, way better!

PR #571, opened 2021-07-08
Some documentation updates

taminomara:
Fixed text style here and there, capitalized Etherscan and NPM, added readme for `hardhat` project and `generate` example.

### DIFF
--- a/ethcontract-common/src/artifact/hardhat.rs
+++ b/ethcontract-common/src/artifact/hardhat.rs
@@ -99,7 +99,7 @@ pub struct HardHatLoader {
 }
 
 impl HardHatLoader {
-    /// Create a new hardhat loader.
+    /// Creates a new hardhat loader.
     pub fn new() -> Self {
         HardHatLoader {
             origin: None,
@@ -110,7 +110,7 @@ impl HardHatLoader {
         }
     }
 
-    /// Create a new hardhat loader and set an override for artifact's origins.
+    /// Creates a new hardhat loader and sets an override for artifact's origins.z
     pub fn with_origin(origin: impl Into<String>) -> Self {
         HardHatLoader {
             origin: Some(origin.into()),
@@ -121,7 +121,7 @@ impl HardHatLoader {
         }
     }
 
-    /// Set new override for artifact's origin. See [`origin`] for more info.
+    /// Sets new override for artifact's origin. See [`origin`] for more info.
     ///
     /// [`origin`]: #structfield.origin
     pub fn origin(mut self, origin: impl Into<String>) -> Self {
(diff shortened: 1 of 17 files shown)


ethcontract-common/src/artifact/hardhat.rs
```diff
@@ -181,27 +181,27 @@ impl HardHatLoader {
         self
     }
 
-    /// Loads an artifact from a JSON value.
+    /// Load an artifact from a JSON value.
```
bh2smith: In this particular case I actually prefer "Loads", and I think (but not completely sure) that "Loads" is actually more correct here. 

Example Scenario
- Object: Car
- Function: Drive on road
- Description: Drives car on road <---- Not "drive on road"

```py 
# drives self on road
def drive_on_road(self: Car):
```

ethcontract-derive/src/lib.rs
```diff
@@ -43,14 +43,14 @@ use syn::{
 /// // HTTP(S) source
 /// contract!("https://my.domain.local/path/to/contract.json")
 ///
-/// // etherscan.io
+/// // Etherscan.io
```
bh2smith: I believe this is a URL... and should remain lower case.

examples/hardhat/hardhat.config.js
```diff
@@ -1,6 +1,6 @@
 require("hardhat-deploy");
 
-const {PK, INFURA_KEY} = process.env;
+const {PK, INFURA_PROJECT_ID} = process.env;
```
bh2smith, line 3: Now this is an interesting change, because it makes so much sense (since that is literally what it is), although I suspect this will take some time for people to adjust to - including myself.

bh2smith (approved):
Approving because its more or less fine, although there are some changes that should probably be reverted (load/loads and one capitalized etherscan)

examples/hardhat/hardhat.config.js
taminomara: To be fair, this change was made just to be consistent with truffle project 😉

ethcontract-common/src/artifact/hardhat.rs
taminomara: You're right, and I was using the wrong form this whole time XD I'll revert this change and update other comments.

taminomara:
Oh by the way, if you see any things about my english that can be improved, feel free to point them out!

ethcontract-common/src/artifact/hardhat.rs
taminomara: Wait, so we have a mix of both in our documentation comments. My inner perfectionist is not very happy about it =)

examples/hardhat/hardhat.config.js
bh2smith: Yes of course and this is why I say it makes so much sense.

taminomara:
@bh2smith I've added more changes, could you please take another look?

vkgnosis (approved):

PR #572, opened 2021-07-09
Enable dependabot updates for hardhat project

taminomara:
I'm not sure if we need this. We don't build hardhat project regularly, because all artifacts are committed to the repo.


vkgnosis (approved):
The config itself looks fine. But I don't know if we need this either.

taminomara:
Well let's merge it and see how it goes. We can always disable it later.

PR #598, opened 2021-08-16
Move convenience utilities for tests to a public module

taminomara:
Instead of re-implementing these small functions in every crate, we could have them here.

### DIFF
--- a/ethcontract-mock/src/lib.rs
+++ b/ethcontract-mock/src/lib.rs
@@ -232,6 +232,7 @@ pub use ethcontract::contract::Signature;
 mod details;
 mod predicate;
 mod range;
+pub mod utils;
 
 #[cfg(test)]
 mod test;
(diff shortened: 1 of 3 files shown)


ethcontract-mock/src/lib.rs
```diff
@@ -232,6 +232,7 @@ pub use ethcontract::contract::Signature;
 mod details;
 mod predicate;
 mod range;
+pub mod utils;
```
nlordell, line 235: Should this maybe be behind a feature?

taminomara: I think it doesn't add too much code size or complexity to warrant a feature. On the other hand, it may grow at some point. I'm not sure about this.

vkgnosis (approved):
Useful enough to be public imo. I don't think a feature is needed unless this added significant compile through large dependencies or the like.

ethcontract-mock/src/lib.rs
nlordell: Lets revisit the feature idea if it adds specific crate dependencies.

nlordell (approved):

PR #600, opened 2021-08-17
Fix bump script for OSX

taminomara:
OSX `sed` is not compatible with GNU `sed`. It has different syntax for `-i` option. See this [stackoverflow] question.

[stackoverflow]: https://stackoverflow.com/questions/4247068/sed-command-with-i-option-failing-on-mac-but-works-on-linux

### DIFF
--- a/ci/bump.sh
+++ b/ci/bump.sh
@@ -65,5 +65,9 @@ function msg {
 msg "Updating Cargo manifests with new version '$version':"
 for manifest in ethcontract*/Cargo.toml; do
 	msg "  - $manifest"
-	sed -i -E -e 's/^((ethcontract-[a-z]+ = \{ )?version) = "[0-9\.]+"/\1 = "'"$version"'"/g' "$manifest"
+	if [[ $(uname) == Darwin ]]; then
+		sed -i '' -E -e 's/^((ethcontract(-[a-z]+)? = \{ )?version) = "[0-9\.]+"/\1 = "'"$version"'"/g' "$manifest"
+	else
+		sed -i -E -e 's/^((ethcontract(-[a-z]+)? = \{ )?version) = "[0-9\.]+"/\1 = "'"$version"'"/g' "$manifest"
+	fi
 done


ci/bump.sh
```diff
@@ -65,5 +65,9 @@ function msg {
 msg "Updating Cargo manifests with new version '$version':"
 for manifest in ethcontract*/Cargo.toml; do
 	msg "  - $manifest"
-	sed -i -E -e 's/^((ethcontract-[a-z]+ = \{ )?version) = "[0-9\.]+"/\1 = "'"$version"'"/g' "$manifest"
+	if [[ $(uname) == Darwin ]]; then
+	  sed -i '' -E -e 's/^((ethcontract(-[a-z]+)? = \{ )?version) = "[0-9\.]+"/\1 = "'"$version"'"/g' "$manifest"
```
nlordell: ubernit: Something seems wrong with the indentation here.

nlordell (approved):
Minor indentation issue I think. Maybe related to tab/spaces?

ci/bump.sh
taminomara: Fixed 👍

ci/bump.sh
```diff
@@ -65,5 +65,9 @@ function msg {
 msg "Updating Cargo manifests with new version '$version':"
 for manifest in ethcontract*/Cargo.toml; do
 	msg "  - $manifest"
-	sed -i -E -e 's/^((ethcontract-[a-z]+ = \{ )?version) = "[0-9\.]+"/\1 = "'"$version"'"/g' "$manifest"
+	if [[ $(uname) == Darwin ]]; then
+		sed -i '' -E -e 's/^((ethcontract(-[a-z]+)? = \{ )?version) = "[0-9\.]+"/\1 = "'"$version"'"/g' "$manifest"
```
taminomara, line 69: Also notice that this regex was changed to update `ethcontract = ...` as well (previously it was `ethcontract-... = ...`). I totally forgot about this change when I was creating this PR 😕

PR #601, opened 2021-08-17
Implement `BatchTransport` for `DynTransport`

taminomara:
This is required to use mock transport in GPv2 code.

### Test Plan
Added test for batches in mock transport. The rest is covered by existing tests.

### DIFF
--- a/ethcontract/src/transport.rs
+++ b/ethcontract/src/transport.rs
@@ -15,27 +15,34 @@ use std::fmt::Debug;
 use std::future::Future;
 use std::sync::Arc;
 use web3::error::Error as Web3Error;
-use web3::{RequestId, Transport};
+use web3::{BatchTransport, RequestId, Transport};
 
 /// Type alias for the output future in for the `DynTransport`'s `Transport`
 /// implementation.
 type BoxedFuture = BoxFuture<'static, Result<Value, Web3Error>>;
+type BoxedBatch = BoxFuture<'static, Result<Vec<Result<Value, Web3Error>>, Web3Error>>;
 
 /// Helper trait that wraps `Transport` trait so it can be used as a trait
 /// object. This trait is implemented for all `Transport`'s.
 trait TransportBoxed: Debug {
     /// Wraps `Transport::prepend`
     fn prepare_boxed(&self, method: &str, params: Vec<Value>) -> (RequestId, Call);
+
     /// Wraps `Transport::send`
     fn send_boxed(&self, id: RequestId, request: Call) -> BoxedFuture;
+
     /// Wraps `Transport::execute`
     fn execute_boxed(&self, method: &str, params: Vec<Value>) -> BoxedFuture;
+
+    /// Wraps `BatchTransport::send_batch`
+    fn send_batch_boxed(&self, requests: Vec<(RequestId, Call)>) -> BoxedBatch;
 }
 
-impl<F, T> TransportBoxed for T
+impl<F, B, T> TransportBoxed for T
 where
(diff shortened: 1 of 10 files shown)


taminomara:
`DynTransport` now requires any wrapped transport to implement `BatchTransport`. I was thinking about making a new struct `DynBatchTransport`, but decided against it, at least for now. I don't want to spend too much time on this, and all transports that we use in our services implement batching anyways. We could split `DynTransport` and `DynBatchTransport` later if there is such need. It's important to decide on this before stabilising the crate though.

ethcontract/src/transport.rs
```diff
@@ -62,10 +73,11 @@ pub struct DynTransport {
 
 impl DynTransport {
     /// Wrap a `Transport` in a `DynTransport`
-    pub fn new<F, T>(inner: T) -> Self
+    pub fn new<F, B, T>(inner: T) -> Self
     where
         F: Future<Output = Result<Value, Web3Error>> + Send + 'static,
-        T: Transport<Out = F> + Sync + Send + 'static,
+        B: Future<Output = Result<Vec<Result<Value, Web3Error>>, Web3Error>> + Send + 'static,
+        T: Transport<Out = F> + BatchTransport<Batch = B> + Sync + Send + 'static,
```
vkgnosis, line 80: The way this is written (and in the generated code before) means that the transport has to be Transport AND BatchTransport whereas before only Transport was required. Maybe this isn't a problem because in practice all Transports are also BatchTransport but I wonder if there is a way to not require this.
Maybe there could be a DynBatchTransport which is like DynTransport but with Batch?
Or DynTransport could only implement BatchTransport if the inner Transport is BatchTransport.

taminomara: > Maybe there could be a DynBatchTransport which is like DynTransport but with Batch?

I was thinking about it, see [above][1].

> Or DynTransport could only implement BatchTransport if the inner Transport is BatchTransport.

This would be the best, but it's not possible because `DynTransport` is not generic.

[1]: https://github.com/gnosis/ethcontract-rs/pull/601#issuecomment-900183894

taminomara: If you think that `DynBatchTransport` should be included in this PR, I can implement it. Maybe someone else have something to add?

vkgnosis: Didn't see your message when I was reviewing. What you say makes sense. It would be nice if there was DynBatchTransport but I don't think it is necessary. Like I said, in practice all transports do implement it. I don't think it is a big deal to breaking change this later either since this is a fairly niche use case.

vkgnosis (approved):
Looks good to me based on the discussion thread. Let's still wait to see what Nick thinks before merging.

taminomara:
So I've tried implementing `DynBatchTransport`, you can see it in #602.

There's now a conversion from `DynBatchTransport` to `DynTransport`. It requires converting `Arc<dyn BatchTransportBoxed>` to `Arc<dyn TransportBoxed>` which in turn requires virtual calls.

To be honest, I'm not happy with either of them.

nlordell:
Personally, I kind of prefer just requiring Batch transport for contract instances, especially considering that the transports we use are batched anyway.

When I first came up with this `DynTransport` thing, it was more of a way to simplify how the generated code was used (i.e. it was just an opaque type without the need for generics). Maybe we should also re-evaluate that decision as it seems to be causing issues :stuck_out_tongue:.

nlordell (approved):

nlordell (approved):

taminomara:
Then we'll go with this implementation for now 😉

PR #608, opened 2021-08-18
Add ability to downcast `DynTransport`

### LINKED ISSUE #551: Add ability to downcast DynWeb3 into conctrete Web3 type.
It would be useful for `DynWeb3` to be able to downcast to a regular `Web3` type with a known transport type. We could use this for instance for:

```rust
let web3: Web3<SomeTransport> = ...;
let instance = MyContract::deployed(&web3).await?;
drop(web3);
let web3: Web3<SomeTransport> = instance.web3().downcast()?;
```

taminomara:
Fix #551.

This PR adds method to downcast `DynTransport` into the wrapped transport type.

Note: we can't move out from `Arc`, so we return reference to the internal transport, which can be cloned. We also can't add methods to `Web3` struct because of orphan rule.


bh2smith (approved):
Very fancy! today I learned about the "orphan rule"

vkgnosis (approved):

PR #614, opened 2021-08-23
More utils (pull useful stuff from cowswap services)

taminomara:


nlordell (approved):
Looks good, my only comment would be to use account 0 from the deterministic Ganache mnemonic. "Alice" is fine though :stuck_out_tongue:.

vkgnosis (approved):

PR #619, opened 2021-08-31
Support `net_version` call

taminomara:
Some of our calls use `net_version` instead of `eth_chainId`, so we need to support both.
### Test Plan
New unittest


bh2smith (approved):
I don't really know what net version means, but this all seems legit.

nlordell:
>what net version means

Its a JSON RPC method part of the Ethereum JSON RPC standard:
https://eth.wiki/json-rpc/API#net_version

nlordell (approved):
Legit

PR #620, opened 2021-08-31
Release 0.16.0

taminomara:
- Support `net_version` call (#619)
- Add ability to downcast `DynTransport` (#608)

### Test Plan

CI


vkgnosis (approved):
PR title is 0.16.0 but actual version is 0.15.2. The latter makes more sense to me anyway but 0.16 would also be fine.

taminomara:
Do we only increment minor version if we break the API?

taminomara:
I'll create another PR with 0.15.2.

PR #623, opened 2021-08-31
Fix handling of functions that return multiple values

taminomara:
When ethcontract determines appropriate rust type for function output, it has a special case for functions that return a single element. Normally, function return type would always be a tuple. With a single return value, the tuple is unwrapped to make code more ergonomic:

- for `function x()` output type is `()`,
- for `function x() returns (T)` output type is `T` (not `(T,)`),
- for `function x() returns (A, B, ...)` output type is `(A, B, ...)`.

We need to account for this conversion and wrap output of any function that returns a single value into an additional tuple.

### Test Plan
New unit tests


vkgnosis (approved):

bh2smith (approved):

bh2smith:
After this, I presume we need another PR with a version bump?

taminomara:
@bh2smith yes, #624.

PR #638, opened 2021-09-22
Add support for 1559 transactions

sunce86:
This PR adds a possibility to build and send the eip1559 type of transactions.
Beside that, `TransactionBuilder` is no longer responsible for providing the default gas price by resolving it from the ethereum node. Gas price is changed according to the rules:

In the past, we had:
```
GasPrice::Standard // use `eth_estaimteGasPrice`
GasPrice::Scaled(..) // use `eth_estaimteGasPrice` and mltiply by factor, wanted to get rid of this
GasPrice::Value(..) // use the specified value.
```

I was more suggesting getting rid of `GasPrice` and instead using `Option<U256>` on the `TransactionBuilder`. In this case, we would have:
```
None // Note that `rust-web3` crate allows `None` values for its transaction requests,
     // and would behave like `GasPrice::Standard` used to
Some(..) // behaves like the `GasPrice::Value(..)` used to
```

With the addition of EIP-1559 gas pricing, I would suggest using `Option<GasPrice>` in the `TrasnactionBuilder` so that:
```
None // Note that `rust-web3` crate allows `None` values for its transaction requests,
     // and would behave like `GasPrice::Standard` used to
Some(GasPrice::Legacy(..)) // behaves like the `GasPrice::Value(..)` used to
Some(GasPrice::Eip1559(..)) // the cool new stuff :D
```

Related PRs:
https://github.com/gnosis/gp-gas-estimation/pull/14
https://github.com/gnosis/gp-transaction-retry/pull/5
https://github.com/gnosis/gp-v2-services/pull/1166
https://github.com/gnosis/gp-v2-services/issues/1056


ethcontract/src/transaction/typed_gas_price.rs
```diff
@@ -0,0 +1,229 @@
+//! Implementation of typed gas price estimation.
+
+use crate::errors::ExecutionError;
+use crate::GasPrice;
```
nlordell: Personally, I'd be OK with getting rid of the `GasPrice` thing that we added a while back. The intention was to "scale the web3 gas estaimte by a factor", but it ended up never really being used or that useful. I imagine removing it would simplify things.

I'm not sure what the others things about it.

sunce86: Yes, I believe only GasPrice::Value variant is used in the code, not sure if other variants have any purpose, expecially since we can call node.gas_price() from the gas price estimator.
Removing it would definitely make code much cleaner.

nlordell: Great! Lets purge it with fire then! :stuck_out_tongue:

ethcontract/src/contract/method.rs
```diff
@@ -227,17 +227,25 @@ impl<T: Transport, R: Tokenize> ViewMethodBuilder<T, R> {
...
+            .unwrap_or_default()
+            .resolve_for_transaction();
         (
             self.m.function,
             CallRequest {
                 from: self.m.tx.from.map(|account| account.address()),
                 to: Some(self.m.tx.to.unwrap_or_default()),
                 gas: self.m.tx.gas,
-                gas_price: self.m.tx.gas_price.and_then(|gas_price| gas_price.value()),
+                gas_price,
```
nlordell: :tada:

ethcontract/src/secret.rs
```diff
@@ -103,6 +103,10 @@ impl Key for &'_ PrivateKey {
     fn address(&self) -> Address {
         self.public_address()
     }
+
+    fn sign_message(&self, message: &[u8]) -> Result<Signature, SigningError> {
+        todo!()
```
nlordell: Ahh, is this is because the recovery value is 0 or 1 for the new transaction types instead 27 or 28? Wouldn't this be as easy as:
```suggestion
        let signature = self.sign(message, None)?;
        Ok(Signature { v: signature.v - 27, ..signature })
```

ethcontract/src/transaction/gas_price.rs
```diff
@@ -1,121 +1,60 @@
...
-    /// Specify a specific gas price to use for the transaction. This will cause
-    /// the transaction `SendFuture` to not query the node for a gas price
-    /// estimation.
-    Value(U256),
+    /// Legacy type of transactions, using single gas price value. Equivalent to sending
+    /// eip1559 transaction with max_fee_per_gas = max_priority_fee_per_gas = gas_price
+    Legacy(U256),
+
+    /// Eip1559 type of transactions, using two values (max_fee_per_gas, max_priority_fee_per_gas)
+    Eip1559((U256, U256)),
```
nlordell: nit: It would be nice to use names so that there are no confusions about the order:
```suggestion
    Eip1559 {
        max_fee_per_gas: U256,
        max_priority_fee_per_gas: U256,
    }
```

ethcontract/src/transaction/gas_price.rs
```diff
@@ -1,121 +1,60 @@
...
-    }
-
-    /// Returns `Some(value)` if the gas price is explicitly specified, `None`
-    /// otherwise.
-    pub fn value(&self) -> Option<U256> {
+    /// Prepares the data for transaction. Returns tuple:
+    /// (gas_price, max_fee_per_gas, max_priority_fee_per_gas, transaction_type)
+    pub fn resolve_for_transaction(
+        &self,
+    ) -> (Option<U256>, Option<U256>, Option<U256>, Option<U64>) {
```
nlordell: Might make sense define a structure for this return value? In general (like :point_up:) I find that these larger tuples with where many fields have the same type make it easy to accidentally get the order of the return values hard, which makes for some hard to find bugs and some extra burden on reviews:

```rust
struct ResolvedTransactionGasPrice {
  pub gas_price: Option<U256>,
  pub max_fee_per_gas: Option<U256>,
  pub max_priority_fee_per_gas: Option<U256>,
  pub transaction_type: Option<U64>,
}
```

ethcontract/src/transaction/gas_price.rs
```diff
@@ -1,121 +1,60 @@
...
     }
-
-    /// Resolves the gas price into a value. Returns a future that resolves once
-    /// the gas price is calculated as this may require contacting the node for
-    /// gas price estimates in the case of `GasPrice::Standard` and
-    /// `GasPrice::Scaled`.
-    pub async fn resolve<T: Transport>(self, web3: &Web3<T>) -> Result<U256, ExecutionError> {
-        let resolved_gas_price = match self {
-            GasPrice::Standard => web3.eth().gas_price().await?,
-            GasPrice::Scaled(factor) => {
```
nlordell, line 56: So happy this is gone :stuck_out_tongue:

ethcontract/src/transaction/gas_price.rs
```diff
@@ -1,121 +1,60 @@
...
-        let gas_price = match self {
-            GasPrice::Standard => None,
-            _ => Some(self.resolve(web3).await?),
-        };
-
-        Ok(gas_price)
-    }
 }
 
 impl Default for GasPrice {
```
nlordell: Maybe a `Default` gas price implementation doesn't make much sense anymore. I'd be OK to remove it, unless its used somehwere.

the `TransactionBuilder` type probably needs `gas_price: Option<GasPrice>` which would estimate the legacy gas price from the node if its not provided (kind of how we fetch the `nonce` if it isn't provided).

nlordell (commented):
Nice! Looks good, just some minor comments about using structs with named fields instead of tuples.

ethcontract/src/secret.rs
sunce86: Not 100% sure, but I expect this function to be erased when I fix the cargo dependencies. For now I leave this comment as unresolved.

ethcontract/src/transaction/gas_price.rs
sunce86: With erasing `GasPrice::Standard` and `GasPrice::Scaled` we are enforcing a user of `TransactionBuilder` to provide gas price or else no gas price will be used for this transaction (resolving of the gas price by asking the node for estimate is removed from `TransactionBuilder'). Is this the desired behavior? I would say yes based on the: https://github.com/gnosis/ethcontract-rs/pull/638#discussion_r716433621

ethcontract/src/transaction/gas_price.rs
sunce86: Agreed. Fixed.

ethcontract/src/transaction/gas_price.rs
sunce86: Agreed. Fixed.

ethcontract/src/transaction/gas_price.rs
nlordell: In the past, we had:
```
GasPrice::Standard // use `eth_estaimteGasPrice`
GasPrice::Scaled(..) // use `eth_estaimteGasPrice` and mltiply by factor, wanted to get rid of this
GasPrice::Value(..) // use the specified value.
```

I was more suggesting getting rid of `GasPrice` and instead using `Option<U256>` on the `TransactionBuilder`. In this case, we would have:
```
None // Note that `rust-web3` crate allows `None` values for its transaction requests,
     // and would behave like `GasPrice::Standard` used to
Some(..) // behaves like the `GasPrice::Value(..)` used to
```

With the addition of EIP-1559 gas pricing, I would suggest using `Option<GasPrice>` in the `TrasnactionBuilder` so that:
```
None // Note that `rust-web3` crate allows `None` values for its transaction requests,
     // and would behave like `GasPrice::Standard` used to
Some(GasPrice::Legacy(..)) // behaves like the `GasPrice::Value(..)` used to
Some(GasPrice::Eip1559(..)) // the cool new stuff :D
```

Because `web3` accepts `None` for the gas price and trasnaction type fields for executing transaction, personal signing transaction, and offline signing transactions, this means we may be able to use use `ResolvedTransactionGasPrice::default()` if `gas_price: Option<GasPrice> == None`. This way we don't require an explicit gas price for every transaction (which would make things a bit more un-ergonomic).

sunce86: Yes, with the last commit (https://github.com/gnosis/ethcontract-rs/pull/638/commits/87b0e64b06723718d408a24d61fec44d11bd3407) I think the implementation is aligned with the above comment. Also, moving last comment to PR description for more visibility.

ethcontract/src/transaction/build.rs
```diff
@@ -36,7 +35,7 @@ impl<T: Transport> TransactionBuilder<T> {
                 build_transaction_request_for_local_signing(
                     self.web3,
                     None,
-                    gas_price,
+                    self.gas_price,
```
nlordell: We might even be able to move this to the `TransactionOptions` as a field now.

sunce86: True :) Moved.

nlordell (commented):
Looks good to me! Just a couple of missing TODOs.

ethcontract-common/Cargo.toml
```diff
@@ -19,4 +19,4 @@ serde_derive = "1.0"
 serde_json = "1.0"
 thiserror = "1.0"
 tiny-keccak = { version = "2.0", features = ["keccak"] }
-web3 = { version = "0.17", default-features = false }
+web3 = { git = "https://github.com/tomusdrw/rust-web3.git" }
```
bh2smith: How long do we expect to have to depend on this git crate? What specific feature does this version have that isn't already available in v0.17?

ethcontract-mock/src/details/mod.rs
```diff
@@ -643,6 +643,7 @@ impl MockTransport {
             root: None,
             logs_bloom: Default::default(),
             transaction_type: None,
+            effective_gas_price: Default::default(), // todo
```
bh2smith: What will this eventually be replaced with?

examples/documentation/src/main.rs
```diff
@@ -5,9 +5,9 @@
...
 //! illustrate what the generated API.
 
-ethcontract::contract!("examples/truffle/build/contracts/DocumentedContract.json",);
-ethcontract::contract!("examples/truffle/build/contracts/SimpleLibrary.json",);
-ethcontract::contract!("examples/truffle/build/contracts/LinkedContract.json",);
-ethcontract::contract!("examples/truffle/build/contracts/IERC20.json",);
+//ethcontract::contract!("examples/truffle/build/contracts/DocumentedContract.json",);
+//ethcontract::contract!("examples/truffle/build/contracts/SimpleLibrary.json",);
+//ethcontract::contract!("examples/truffle/build/contracts/LinkedContract.json",);
+//ethcontract::contract!("examples/truffle/build/contracts/IERC20.json",);
```
bh2smith: I'm guessing this is temporary...

ethcontract-common/Cargo.toml
sunce86: We need this one to be released:
https://github.com/tomusdrw/rust-web3/commit/2e2d1dfa906aa2dff6312b97d0d5ba804e9c8c6e
ETA: Unknown :)

ethcontract/src/secret.rs
sunce86: Implemented.

ethcontract-mock/src/details/mod.rs
sunce86: Populated with legacy gas price. Might be changed in the next PR for MockTransport update.

examples/documentation/src/main.rs
sunce86: Reverted.

ethcontract/src/secret.rs
```diff
@@ -100,6 +100,19 @@ impl Key for &'_ PrivateKey {
         Ok(Signature { v, r, s })
```
nlordell: nit: Could you refactor `sign` in function of `sign_message`?

nlordell (approved):

ethcontract/src/secret.rs
sunce86: Done.

bh2smith (approved):

bh2smith:
Looks good here - the only remaining errors appear to be from clippy (remove unnecessary clone).

PR #658, opened 2021-11-03
Fix missing fmt argument in error message

dtolnay:
Without this, the error message would literally be "root path '{}' is not absolute" with curly braces in it instead of the root path.

### Test Plan
`cargo check --manifest-path ethcontract-generate/Cargo.toml`


nlordell (approved):
Nice find! Thanks for the PR :smile:

nlordell:
Also, just wanted to say thanks for all your amazing work! Off the top of my head we use `anyhow`, `thiserror`, `syn`, `quote`, `serde(-.*)?` and which are huge center pieces in the Rust crate ecosystem.

bh2smith (approved):
Cool beans! Nice catch.

vkgnosis (approved):

PR #667, opened 2021-11-18
Add nonce() to ViewMethodBuilder

hyche:
More convenient to use instead of building from scratch using
MethodBuilder

### Test Plan
no new test added just execute 
`cargo test`

### DIFF
--- a/ethcontract/src/contract/method.rs
+++ b/ethcontract/src/contract/method.rs
@@ -203,8 +203,8 @@ impl<T: Transport, R: Tokenize> ViewMethodBuilder<T, R> {
         self
     }
 
-    /// Specify the nonce for the transation, if not specified will use the
-    /// current transaction count for the signing account.
+    /// Specify the block height for the call, if not specified then latest
+    /// mined block will be used.
     pub fn block(mut self, value: BlockId) -> Self {
         self.block = Some(value);
         self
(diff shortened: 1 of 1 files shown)


nlordell (approved):
LGTM!

ethcontract/src/contract/method.rs
```diff
@@ -196,6 +196,21 @@ impl<T: Transport, R: Tokenize> ViewMethodBuilder<T, R> {
...
+        self
+    }
+
+    /// Specify the number of confirmations to wait for when confirming the
+    /// transaction, if not specified will wait for the transaction to be mined
+    /// without any extra confirmations.
+    pub fn confirmations(mut self, value: usize) -> Self {
+        self.m = self.m.confirmations(value);
+        self
+    }
```
nlordell: I just remembered, this doesn't do anything for view methods AFAIK (because they don't get mined so it doesn't make sense to wait for block confirmations).

I agree with the `nonce()` helper method though!

nlordell (changes requested):
Requesting to remove the `confirmations()` method since I'm not sure what it can be used for.

If i'm missing something let me know!

ethcontract/src/contract/method.rs
hyche: Because the method returns a state that might be changed in the latest blocks, and when reorg happened the state will be reverted, which makes different views of a data for each call. I think confirmations() would prevent that.

nlordell: But the `ViewMethodBuilder` doesn't make use of `confirmations` in the `call()`.

I think in order to achieve that at the moment, you would need to specify the `block_number` for calling (and not the `confirmations`).

hyche: I see thanks for the clarification, fixed the PR.

nlordell (approved):

PR #700, opened 2022-01-17
Put GasPrice module under a feature

sunce86:
~~Added conversion from EstimatedGasPrice to GasPrice.
This conversion is used in multiple places in gp-v2-services repo, thus this PR will reduce code duplication.~~

~~Planning to merge regardless of clippy errors (probably consequence of the Rust outdated version)~~

This PR puts GasPrice under a feature, so only its definition could be imported into gp-gas-estimation repo

### DIFF
--- a/ethcontract/Cargo.toml
+++ b/ethcontract/Cargo.toml
@@ -33,6 +33,7 @@ ipc-tokio = ["web3/ipc-tokio"]
 arrayvec = "0.7"
 ethcontract-common = { version = "0.15.4", path = "../ethcontract-common" }
 ethcontract-derive = { version = "0.15.4", path = "../ethcontract-derive", optional = true}
+gas-estimation = { git = "https://github.com/gnosis/gp-gas-estimation.git", tag = "v0.4.0"}
 futures = "0.3"
 futures-timer = "3.0"
 hex = "0.4"
--- a/ethcontract/Cargo.toml
+++ b/ethcontract/Cargo.toml
@@ -17,7 +17,8 @@ to Ethereum smart contracts.
 name = "ethcontract"
 
 [features]
-default = ["derive", "http-tls", "ws-tls-tokio"]
+default = ["derive", "http-tls", "ws-tls-tokio", "gas_price"]
+gas_price = ["primitive-types", "web3"]
 derive = ["ethcontract-derive"]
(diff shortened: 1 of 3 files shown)


ethcontract/Cargo.toml
```diff
@@ -33,6 +33,7 @@ ipc-tokio = ["web3/ipc-tokio"]
 arrayvec = "0.7"
 ethcontract-common = { version = "0.15.4", path = "../ethcontract-common" }
 ethcontract-derive = { version = "0.15.4", path = "../ethcontract-derive", optional = true}
+gas-estimation = { git = "https://github.com/gnosis/gp-gas-estimation.git", tag = "v0.4.0"}
```
nlordell: Not a huge fan of pulling in a `git` dependency here (this means that we can't publish to [crates.io](https://crates.io) for example).

Maybe, instead of adding the conversion here, we can add it to `gp-gas-estimation` behind a feature (to avoid building additional dependencies that aren't needed)?

nlordell:
In general I like it! I think it would be better to move this to the `gas-estimation` crate however behind a feature so that we can continue to `cargo publish` this crate once.

ethcontract/Cargo.toml
sunce86: Was hoping for this kind of a suggestion :)
So, just to confirm, put `gas_price` module in `ethcontract-rs` under feature, then import that feature in `gp-gas-estimation`?
`gp-v2-services` already uses `EstimatedGasPrice` (where conversion will be added) so no need to put only the conversion inside a feature I guess

nlordell: 🙈 
Sorry I didn't see the reply. What I meant was making `ethcontract` optional in `gp-gas-estimation` so that it only pulls in the dependency if you:
```toml
gas-estimation = { version = "*", features = ["ethcontract"] }
```

With my suggestion, there were no changes required in `ethcontract`.

I think `ethcontract` needs `gas_price` module internally so I don't know if the feature would work here like this.

sunce86:
Closing in favor of: https://github.com/gnosis/gp-v2-services/pull/1579
Context here: https://github.com/gnosis/gp-gas-estimation/pull/21#issuecomment-1015279449

PR #721, opened 2022-02-17
Make ethcontract-generate curl dependency optional

vkgnosis:
I noticed that in services we do not make use of this feature but every ci run curl-sys is recompiled which caused ethcontract to be compiled. I'm not sure why that happens but either way we can make this optional.
Not entirely sure if this is worth it. Would also help if for whatever reason you don't have curl-sys because that is a c dependency which might not exist on your platform.

### Test Plan
CI


nlordell (approved):
Nice catch.

PR #723, opened 2022-02-17
Support eip1559

vkgnosis:
I decided to not wait and try this myself.
This is an update of #638 fixing the merge conflicts. There were almost no code changes compared to the original PR.

Closes https://github.com/gnosis/ethcontract-rs/pull/638

### Test Plan
CI

### DIFF
--- a/ethcontract-mock/src/details/mod.rs
+++ b/ethcontract-mock/src/details/mod.rs
@@ -643,7 +643,7 @@ impl MockTransport {
             root: None,
             logs_bloom: Default::default(),
             transaction_type: None,
-            effective_gas_price: Some(U256::from(1)),
+            effective_gas_price: Some(tx.gas_price),
         };
 
         state.receipts.insert(tx.hash, receipt);
--- a/ethcontract-mock/src/details/sign.rs
+++ b/ethcontract-mock/src/details/sign.rs
@@ -86,5 +86,8 @@ pub fn verify(raw_tx: &[u8], node_chain_id: u64) -> Transaction {
         value,
         data,
         hash: signing::keccak256(raw_tx).into(),
+        transaction_type: 0,
+        max_fee_per_gas: Default::default(),
+        max_priority_fee_per_gas: Default::default(),
     }
 }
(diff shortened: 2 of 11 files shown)


nlordell:
I did `git diff origin/eip1559 origin/vk-eip1559` and it looks like some clean-ups didn't make it over for some reason. For example:

```diff
diff --git a/ethcontract-common/src/bytecode.rs b/ethcontract-common/src/bytecode.rs
index db22786..707f861 100644
--- a/ethcontract-common/src/bytecode.rs
+++ b/ethcontract-common/src/bytecode.rs
@@ -60,7 +60,9 @@ impl Bytecode {
         S: AsRef<str>,
     {
         let name = name.as_ref();
-        assert!(!(name.len() > 38), "invalid library name for linking");
+        if name.len() > 38 {
+            panic!("invalid library name for linking");
+        }
 
         // NOTE(nlordell): solc linking works by string search and replace of
         //   '__$name__..__' with the library address; see generated bytecode for
```

Otherwise looks good.

vkgnosis:
I noticed that the branch (but not the PR) included some of these changes where `if X then panic` had been converted to `assert !X`. These don't trigger any warnings so they are not required and as far as I could cell they didn't change logic.
If we want them then let's make another PR because it will be the same amount of work either way and it isn't required for this one.

nlordell:
> If we want them then let's make another PR because it will be the same amount of work either way and it isn't required for this one.

No-no I think its fine. If clippy doesn't complain I don't think its worth changing. Just pointing it out as I was finishing looking over the changes.

The only one that is bothering me a bit (but obviously not worth blocking the PR for is this one:
```diff
     #[test]
     #[cfg_attr(debug_assertions, should_panic)]
     fn div_euclid_overflow() {
-        I256::MIN.div_euclid(-I256::one());
+        // We only attempt to print here because of the must_use
+        // error: unused return value of `int::I256::div_euclid` that must be used
+        println!("{}", I256::MIN.div_euclid(-I256::one()));
     }
```

Where it should be:
```
let _ = I256::MIN.div_euclid(-I256::one());
```

nlordell (approved):

PR #725, opened 2022-02-18
Pass cargo registry token through env variable

vkgnosis:
This is how it is documented https://doc.rust-lang.org/cargo/reference/config.html?highlight=CARGO_REGISTRY_TOKEN#credentials .

Context is that the release job failed https://github.com/gnosis/ethcontract-rs/runs/5247111798 .

### Test Plan
Re-run job afterwards

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -5,6 +5,8 @@ on:
 jobs:
   release:
     runs-on: ubuntu-latest
+    env:
+      CARGO_REGISTRY_TOKEN: ${{ secrets.CARGO_TOKEN }}
     steps:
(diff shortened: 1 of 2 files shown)


nlordell:
Are you sure this is needed and not an issue with an oudated token? https://github.com/gnosis/ethcontract-rs/runs/5247460988?check_suite_focus=true seems to have passed and deployed a new version.

vkgnosis:
I thought this was the issue at first but it it wasn't the reason. Although it still seems cleaner to pass the token by env var.

nlordell (approved):

nlordell:
> Although it still seems cleaner to pass the token by env var

Good point.

PR #759, opened 2022-04-27
deploy with address

sunce86:
This PR adds possibility to deploy new contracts to the network using the custom contract address instead of using the internal counter for automatic address generation.

### DIFF
--- a/ethcontract-mock/src/details/mod.rs
+++ b/ethcontract-mock/src/details/mod.rs
@@ -102,7 +102,8 @@ impl MockTransport {
                 .contracts
                 .insert(address, Contract::new(address, abi))
                 .is_none(),
-            "replacing contract at address {address:?}",
+            "replacing contract at address {:?}",
+            address
         );
     }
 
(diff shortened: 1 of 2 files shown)


nlordell (approved):
LGTM

ethcontract-mock/src/details/mod.rs
```diff
@@ -93,6 +93,13 @@ impl MockTransport {
         address
     }
 
+    /// Deploys a new contract with the given ABI and address
+    pub fn deploy_with_address(&self, abi: &Abi, address: Address) {
+        let mut state = self.state.lock().unwrap();
+
+        state.contracts.insert(address, Contract::new(address, abi));
```
nlordell: Maybe we can assert that we aren't replacing an existing contract?

sunce86: Do you mean adding a simple assert like this:
`assert!(!state.contracts.contains_key(&address));`

vkgnosis (approved):

ethcontract-mock/src/details/mod.rs
nlordell: Also, `insert` returns the previous value, so you can also:
```suggestion
        assert!(
            state.contracts.insert(address, Contract::new(address, abi)).is_none(),
            "replacing contract at address {address:?}",
        );
```

sunce86: Added.

PR #781, opened 2022-06-13
Use Path.join() to create paths instead of Url.path(), fixes #729

### LINKED ISSUE #729: Problem with proc macro contract! (on Windows)
Hi !
I'm recently working with Eth.. I downloaded your sources and wanted to see how the examples work, but I ran into the fact that this macro cannot find files with ABI (The filename, directory name, or volume label syntax is incorrect. (os error 123)) . I tried replacing "/" with "\\" but that didn't help. Can you tell me what I'm doing wrong? Thanks ! )

From under WSL the code is compiled.
image
image

Jaxkr:
We can't assume that the path section of a file:// URL is a valid
operating system path, so instead we join the root and the source using
Rust's std::path API. This still handles absolute paths the same.

This commit resolves issues where we couldn't use ethcontract-derive
on Windows (e.g. gnosis/ethcontract-rs#729)

Fixes #729

### DIFF
--- a/ethcontract-generate/src/source.rs
+++ b/ethcontract-generate/src/source.rs
@@ -110,7 +110,7 @@ impl Source {
         let url = base.join(source.as_ref())?;
 
         match url.scheme() {
-            "file" => Ok(Source::local(url.path())),
+            "file" => Ok(Source::local(root.join(source))),
             #[cfg(feature = "http")]
             "http" | "https" => match url.host_str() {
                 Some("etherscan.io") => Source::etherscan(


nlordell (approved):
LGTM.

vkgnosis (approved):

nlordell:
Looks like some new clippy lints are causing CI to fail. Created https://github.com/gnosis/ethcontract-rs/pull/782 to try and fix them.

vkgnosis:
CI succeeds on the stable which is what we mostly care about anyway.

PR #817, opened 2022-09-04
Ignore clippy type complexity errors in build contracts

josojo:
Smart contracts can have functions or events with a higher type complexity than allowed by clippy. Hence, we need to add the allow statement.

### DIFF
--- a/ethcontract-generate/src/generate.rs
+++ b/ethcontract-generate/src/generate.rs
@@ -131,6 +131,7 @@ fn expand_contract(cx: &Context) -> Result<TokenStream> {
 
     Ok(quote! {
         #[allow(dead_code)]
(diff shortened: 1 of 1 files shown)


ethcontract-generate/src/generate.rs
```diff
@@ -131,6 +131,7 @@ fn expand_contract(cx: &Context) -> Result<TokenStream> {
 
     Ok(quote! {
         #[allow(dead_code)]
+        #[allow(clippy::type_complexity)]
```
nlordell: ubernit:
```suggestion
        #[allow(dead_code, clippy::type_complexity)]
```

nlordell (approved):

vkgnosis (approved):

PR #842, opened 2022-10-10
Promote error to Revert even if decoding the reason fails

sunce86:
Return `Revert` status even if decoding the reason failed.

Should fix the error https://cowservices.slack.com/archives/C0371SB243E/p1665417220790089

This error appears when `eth_call` batch request is sent for `TokenOwnerFinder`: https://github.com/cowprotocol/services/blob/7f2abc427f8627bc712b54ae61977f43c3159a16/crates/shared/src/bad_token/token_owner_finder.rs#L231-L233

Nethermind node asnwers with error message `"Reverted 0x416d6f756e74206d757374206265206c657373207468616e20746f74616c207265666c656374696f6e73"` for one of the requests, which is detected as reverted, but fails to decode `416d6f756e74206d757374206265206c657373207468616e20746f74616c207265666c656374696f6e73` to a reason. This causes this error not to be promoted to `ExecutionError::Revert` but instead to `ExecutionError::Web3` which causes `TokenOwnerFinder` to yield error.

Note: I've spent quite some time to debug why the decoding of the message is failing, but failed to figure out. Tried also to investigate latest changes in Nethermind but with no success. If anyone else is willing to take a look, that would be awesome.

### DIFF
--- a/ethcontract/src/errors/nethermind.rs
+++ b/ethcontract/src/errors/nethermind.rs
@@ -1,5 +1,5 @@
-//! This module implements Parity specific error decoding in order to try and
-//! provide more accurate errors from Parity nodes.
+//! This module implements Nethermind specific error decoding in order to try and
+//! provide more accurate errors from Nethermind nodes.
 
 use crate::errors::{revert, ExecutionError};
 use jsonrpc_core::Error as JsonrpcError;
@@ -9,7 +9,7 @@ const REVERTED: &str = "Reverted 0x";
 /// Invalid op-code error discriminant.
 const INVALID: &str = "Bad instruction";
 
-/// Tries to get a more accurate error from a generic Parity JSON RPC error.
+/// Tries to get a more accurate error from a generic Nethermind JSON RPC error.
 /// Returns `None` when a more accurate error cannot be determined.
 pub fn get_encoded_error(err: &JsonrpcError) -> Option<ExecutionError> {
     let message = get_error_message(err)?;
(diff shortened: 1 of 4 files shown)


vkgnosis:
You are changing the implementation of the parity error detector but the we are trying to decode a nethermind error. I don't think the error format is standardized in any way so it is not surprising that the parity detector cannot handle it and changing this detector seems like the wrong way to go.

Maybe a better is to change https://github.com/gnosis/ethcontract-rs/blob/05bff3fb254dbd0cf6b198b3d4a547d9403acdf4/ethcontract/src/errors.rs#L133 to also try a new `nethermind::get_encoded_error` which could decode this way. We can start with that.
Then it would be still be nice to extract the encoded revert reason for which we'd have to figure out how Nethermind encodes these bytes. We could ask the Nethermind developers about that.

vkgnosis (approved):
This is ok to me now but would wait for Nic's review before merging. One thing I am still wondering is if we have ever successfully decoded a revert reason from nethermind. Maybe the decode_reason function just doesn't apply to it at all.

sunce86:
Maybe we never used Nethermind node with the `tracing` feature. In that case we would get `Reverted 0x` every time and that would never enter the decoding part.

nlordell (approved):

nlordell:
Looks like the CI is failing because of a Rinkeby example test. Feel free to disable it for now (you can just comment them out).

https://github.com/gnosis/ethcontract-rs/blob/05bff3fb254dbd0cf6b198b3d4a547d9403acdf4/.github/workflows/pull-request.yaml#L64-L68

ethcontract/src/errors/nethermind.rs
```diff
@@ -0,0 +1,110 @@
...
+            {
+                Some(reason) => return Some(ExecutionError::Revert(Some(reason))),
+                None => return Some(ExecutionError::Revert(None)),
+            }
+        }
+    } else if message.starts_with(INVALID) {
+        return Some(ExecutionError::InvalidOpcode);
+    }
+
+    None
```
nlordell, line 36: We can also check:

```rust
if err.message == "VM execution error" {
    return Some(ExecutionError::Revert(None));
}
```

As a catch-all (in case we missed some other error messages - like what happens when a contract reverts without data?).

sunce86: I don't see a reason why not to add it. Done.

sunce86:
> Looks like the CI is failing because of a Rinkeby example test. Feel free to disable it for now (you can just comment them out).

I've commented out the rinkeby line but nothing changed. 
Merging since the CI errors don't seem too frightening :)

PR #860, opened 2022-11-07
Allow large_enum_variants

### LINKED ISSUE #854: Rust compiler complaints about Event enum implementation: Different data size
If one has events of very different size in one contract, then the built rust files of a contract contains clippy warnings like these:
```
error: large size difference between variants
   --> /Users/alexherrmann/cow/services/target/debug/build/contracts-f9e055253a6107d1/out/CoWSwapOnchainOrders.rs:421:9
    |
421 |         OrderPlacement(self::event_data::OrderPlacement),
    |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ this variant is 376 bytes
    |
    = note: `-D clippy::large-enum-variant` implied by `-D warnings`
note: and the second-largest variant is 32 bytes:
   --> /Users/alexherrmann/cow/services/target/debug/build/contracts-f9e055253a6107d1/out/CoWSwapOnchainOrders.rs:420:9
    |
420 |         OrderInvalidation(self::event_data::OrderInvalidation),
    |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#large_enum_variant
help: consider boxing the large fields to reduce the total size of the enum
    |
421 |         OrderPlacement(Box<self::event_data::OrderPlacement>),
    |                        ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

```
Is it okay, if I just insert a a comment in these files that we allow `large-enum-variant`. Otherwise, I guess this would be a bigger change that costs quite some time

josojo:
Fixes [#issue](https://github.com/gnosis/ethcontract-rs/issues/854)

This is just the most straightforward thing to do. I think at some point in time, it might make sense to impl the suggestion from the compiler. But I don't wanna block the ethflow project by it.

### Test Plan
non
(If I add this clippy allow it in my Contract.rs of the ethflow contract, I can compile without clippy warnings)

### DIFF
--- a/ethcontract-mock/Cargo.toml
+++ b/ethcontract-mock/Cargo.toml
@@ -14,7 +14,7 @@ Tools for mocking ethereum contracts.
 [dependencies]
 ethcontract = { version = "0.22.0", path = "../ethcontract", default-features = false, features = ["derive"] }
 hex = "0.4"
-mockall = "0.11"
+mockall = "0.11.2"
 rlp = "0.5"
 predicates = "2.0"
 
--- a/ethcontract-mock/Cargo.toml
+++ b/ethcontract-mock/Cargo.toml
@@ -14,7 +14,7 @@ Tools for mocking ethereum contracts.
 [dependencies]
 ethcontract = { version = "0.22.0", path = "../ethcontract", default-features = false, features = ["derive"] }
 hex = "0.4"
-mockall = "0.11.2"
+mockall = "0.11"
 rlp = "0.5"
 predicates = "2.0"
 
--- a/ethcontract-generate/src/generate.rs
+++ b/ethcontract-generate/src/generate.rs
@@ -130,7 +130,7 @@ fn expand_contract(cx: &Context) -> Result<TokenStream> {
     let events = events::expand(cx)?;
 
     Ok(quote! {
-        #[allow(dead_code, clippy::type_complexity)]
+        #[allow(dead_code, clippy::type_complexity, clippy::large_enum_variant)]
         #vis mod #contract_mod {
             #[rustfmt::skip]
             use #runtime_crate as ethcontract;


vkgnosis (approved):
I think this is the correct fix.

MartinquaXD (approved):

josojo:
I can not really fix this issue here:
If I run locally:

```
rustup install 1.56
rustup default 1.56
rm Cargo.lock
cargo build
```
Then I can see in nvim Cargo.lock  

```
[[package]]
name = "ethcontract-mock"
version = "0.22.0"
dependencies = [
 "ethcontract",
 "ethcontract-derive",
 "hex",
 "mockall",
 "predicates",
 "rlp",
 "tokio",
]
```
with
```
[[package]]
name = "mockall"
version = "0.11.3"
source = "registry+https://github.com/rust-lang/crates.io-index"
checksum = "50e4a1c770583dac7ab5e2f6c139153b783a53a1bbee9729613f193e59828326"
dependencies = [
 "cfg-if",
 "downcast",
 "fragile",
 "lazy_static",
 "mockall_derive",
 "predicates",
 "predicates-tree",
]


```
So it still has the wrong mockall version. Why is that? I think its because it still depends on the official ethcontract-mock version. 
How do it test this properly before releasing?

vkgnosis:
What you have specified in Cargo.toml is that mockall can be anything compatible with 0.11 . This is all 0.11.X versions. I think you are trying to specify a maximum mockall version which has different syntax which I don't remember from the top of my head but should be explained in the Cargo book.

Anyway, the better approach is going to be to properly update dependencies and rustc version first.

ethcontract-mock/Cargo.toml
```diff
@@ -14,7 +14,7 @@ Tools for mocking ethereum contracts.
 [dependencies]
 ethcontract = { version = "0.22.0", path = "../ethcontract", default-features = false, features = ["derive"] }
 hex = "0.4"
-mockall = "0.11"
+mockall = "0.11.2"
```
nlordell: This seems like an unrelated change... Is there a reason we need at least `0.11.2`?

nlordell:
I also think that you should update the rustc version to `1.60.0` in `.github/workflows/pull-request.yaml` and `README.md` instead of forcing a specific `mockall` version.

josojo:
>I also think that you should update the rustc version to 1.60.0 in .github/workflows/pull-request.yaml and README.md instead of forcing a specific mockall version.

I only did this, since locally `clippy --all-features` indicated that I would have to also update several other repos and I wanted to save time. Thanks to @vkgnosis who pointed out, that `clippy --all-featueres` never was supped to pass.

nlordell:
Disabled the branch protection rules for 1.56.1. You might need to trick GitHub into re-running CI.

PR #863, opened 2022-11-10
Able to process foundry ABI

josojo:
Foundry and hardhat have a very similar ABI format, but its not exactly the same. The bytecode in foundry is map, instead of just a string. I think the ABI are similar enough, such that we don't have provide another loader format to the existing ones of truffle/hardhat, but can just parse the ByteCode differently. 

Link to [hardhat ABI](https://raw.githubusercontent.com/cowprotocol/ethflowcontract/hardhatversion/build/artifacts/src/CoWSwapEthFlow.sol/CoWSwapEthFlow.json): 
Link to [foundry ABI](https://raw.githubusercontent.com/cowprotocol/ethflowcontract/v0.0.0-rc.1-artifacts/artifacts/CoWSwapEthFlow.sol/CoWSwapEthFlow.json)

### DIFF
--- a/ethcontract-common/src/bytecode.rs
+++ b/ethcontract-common/src/bytecode.rs
@@ -24,7 +24,7 @@ impl Bytecode {
             return Ok(Bytecode::default());
         }
 
-        let s = if s.chars().next().unwrap() == '{' {
+        let s = if s.starts_with('{') {
             #[derive(Deserialize)]
             struct Map {
                 object: String,
--- a/ethcontract-common/src/bytecode.rs
+++ b/ethcontract-common/src/bytecode.rs
@@ -17,12 +17,23 @@ use web3::types::{Address, Bytes};
 pub struct Bytecode(String);
 
 impl Bytecode {
-    /// Reads hex bytecode representation from a string slice.
+    /// Reads hex bytecode representation from a string or map slice.
     pub fn from_hex_str(s: &str) -> Result<Self, BytecodeError> {
         if s.is_empty() {
             // special case where we have an empty string byte code.
             return Ok(Bytecode::default());
         }
+
+        let s = if s.chars().next().unwrap() == '{' {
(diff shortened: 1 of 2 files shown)


ethcontract-common/src/bytecode.rs
```diff
@@ -17,20 +17,31 @@ use web3::types::{Address, Bytes};
...
-    /// Reads hex bytecode representation from a string slice.
+    /// Reads hex bytecode representation from a string or map slice.
     pub fn from_hex_str(s: &str) -> Result<Self, BytecodeError> {
         if s.is_empty() {
             // special case where we have an empty string byte code.
             return Ok(Bytecode::default());
         }
 
+        let s = if s.starts_with('{') {
+            #[derive(Deserialize)]
```
vkgnosis, line 28: This needs some explanation in the code like what you put in the PR description. Should explain why we have this if, what it does, and if possible link to the foundry documentation of their output (not sure this exists).

ethcontract-common/src/bytecode.rs
```diff
@@ -17,20 +17,31 @@ use web3::types::{Address, Bytes};
...
 impl Bytecode {
-    /// Reads hex bytecode representation from a string slice.
+    /// Reads hex bytecode representation from a string or map slice.
     pub fn from_hex_str(s: &str) -> Result<Self, BytecodeError> {
         if s.is_empty() {
             // special case where we have an empty string byte code.
             return Ok(Bytecode::default());
         }
 
+        let s = if s.starts_with('{') {
```
nlordell, line 27: This seems wrong to me. Specifically the Foundry artifact output uses a map and not a string for bytecode (in fact, the actual bytecode in the map has the same format that this type expects).

Maybe the solution is to change the `Hardhat` artifact type to accept `ByteCode | { object: Bytecode }` for its bytecode field (and not change the `Bytecode` parsing logic itself)?

PR #964, opened 2023-11-15
Reduce minor compilation warnings

squadgazzz:
Eliminates minor compilation warnings while using the generated smart contracts:
> warning: accessing first element with `log . topics.get(0)`
 --> /Users/squadgazzz/RustroverProjects/services/target/debug/build/contracts-3ef65e6ec386b1a7/out/IUniswapV3Factory.rs:1:29696
  |
1 | ...ntract :: errors :: ExecutionError > { let standard_event = log . topics . get (0) . copied () . map (| topic | match topic { self :: ethcontract :: H256 ([198 , 106 , 63 , 223...
  |                                                                ^^^^^^^^^^^^^^^^^^^^^^ help: try: `log . topics.first()`
  |
  = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#get_first

### DIFF
--- a/ethcontract/src/errors/nethermind.rs
+++ b/ethcontract/src/errors/nethermind.rs
@@ -46,6 +46,7 @@ fn get_error_message(err: &JsonrpcError) -> &str {
         .unwrap_or(&err.message)
 }
 
+#[allow(unused)]
 #[cfg(test)]
 pub use tests::*;
 
--- a/ethcontract-common/Cargo.toml
+++ b/ethcontract-common/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "ethcontract-common"
-version = "0.25.2"
+version = "0.25.3"
 authors = ["Gnosis developers <developers@gnosis.io>"]
 edition = "2021"
 license = "MIT OR Apache-2.0"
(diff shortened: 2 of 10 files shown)


ethcontract/src/errors/nethermind.rs
```diff
@@ -46,6 +46,7 @@ fn get_error_message(err: &JsonrpcError) -> &str {
         .unwrap_or(&err.message)
 }
 
+#[allow(unused)]
```
MartinquaXD, line 49: Are these `allowed(unused)` actually needed?
I believe we are exporting these types from the library `clippy` shouldn't care about them being used in the crate code.

MartinquaXD (commented):
There is not much development happening in this repo. That's why I would suggest also bumping the crate's version number in this PR so we can immediately import the fixed version in the services repo.

ethcontract/src/errors/nethermind.rs
squadgazzz: But for some reason, `clippy` shows unused warnings, probably because I use it with the `--all-targets` flag.

MartinquaXD (approved):

PR #967, opened 2024-02-19
Replace `net_version` with `eth_chainId`

### LINKED ISSUE #966: Replace use of `net_version` with `eth_chainId`
These calls are identical but some RPCs might not expose the `net` module by default whereas `eth` always gets exposed. Since we only use it for `net_version` this would be an easy win to improve compatibility with nodes.

MartinquaXD:
Fixes #966

* replaces `net_version` calls with `eth_chainId` because that is supported by all nodes by default whereas the `net` functions might not be.
* bumps the version so we can immediately start using this improvement in the services repo
* gets rid of unnecessary imports to avoid clippy warnings on nightly

### Test Plan
Adjusted existing tests

### DIFF
--- a/examples/examples/revert.rs
+++ b/examples/examples/revert.rs
@@ -29,7 +29,7 @@ async fn main() {
     let error = result_0.unwrap_err().inner;
     assert!(matches!(
         error,
-        ExecutionError::Revert(Some(reason)) if reason == "reason"
+        ExecutionError::Revert(Some(reason)) if reason == "revert: reason"
     ));
 
     let error = result_1.unwrap_err().inner;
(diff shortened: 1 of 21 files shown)


examples/examples/revert.rs
```diff
@@ -29,7 +29,7 @@ async fn main() {
     let error = result_0.unwrap_err().inner;
     assert!(matches!(
         error,
-        ExecutionError::Revert(Some(reason)) if reason == "reason"
+        ExecutionError::Revert(Some(reason)) if reason == "revert: reason"
```
MartinquaXD, line 32: Verified with a [test PR](https://github.com/cowprotocol/ethcontract-rs/pull/968) that this change is needed by now even without switching to `eth_chainId`.

fleupold (approved):

PR #969, opened 2024-04-19
Bump crates to the latest versions

mateo-mro:
Bump crates to the latest versions

### DIFF
--- a/.github/workflows/pull-request.yaml
+++ b/.github/workflows/pull-request.yaml
@@ -8,7 +8,7 @@ jobs:
     strategy:
       matrix:
         include:
-          - rust: 1.70.0
+          - rust: 1.77.2
             examples: false
             continue-on-error: false
           - rust: stable
--- a/ethcontract-mock/Cargo.toml
+++ b/ethcontract-mock/Cargo.toml
@@ -14,10 +14,10 @@ Tools for mocking ethereum contracts.
 [dependencies]
 ethcontract = { version = "0.25.6", path = "../ethcontract", default-features = false, features = ["derive"] }
 hex = "0.4"
-mockall = "0.11"
+mockall = "0.12.1"
 rlp = "0.5"
 predicates = "3.0"
 
 [dev-dependencies]
-tokio = { version = "1.6", features = ["macros", "rt"] }
+tokio = { version = "1.37.0", features = ["macros", "rt"] }
 ethcontract-derive = { version = "0.25.6", path = "../ethcontract-derive", default-features = false }
--- a/ethcontract/Cargo.toml
+++ b/ethcontract/Cargo.toml
@@ -32,26 +32,26 @@ ws-tls-tokio = ["web3/ws-tls-tokio"]
 ws-tokio = ["web3/ws-tokio"]
 
 [dependencies]
-aws-config = { version = "0.55", optional = true }
-aws-sdk-kms = { version = "0.28", optional = true }
+aws-config = { version = "1.2.0", features = ["behavior-version-latest"], optional = true }
+aws-sdk-kms = { version = "1.21.0", optional = true }
 arrayvec = "0.7"
 ethcontract-common = { version = "0.25.6", path = "../ethcontract-common" }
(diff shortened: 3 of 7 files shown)


.github/workflows/pull-request.yaml
```diff
@@ -8,7 +8,7 @@ jobs:
     strategy:
       matrix:
         include:
-          - rust: 1.70.0
+          - rust: 1.77.2
```
MartinquaXD, line 11: According to the readme the crate has a minimum supported rust version of 1.64 but I guess that ship has already sailed. 😅 
Is there a reason this needed to be updated? BTW this exact version is already covered by `stable` 3 lines down.

ethcontract-common/Cargo.toml
```diff
@@ -13,10 +13,10 @@ Common types for ethcontract-rs runtime and proc macro.
 
 [dependencies]
 ethabi = "18.0"
-hex = "0.4"
-serde = "1.0"
-serde_derive = "1.0"
-serde_json = "1.0"
-thiserror = "1.0"
-tiny-keccak = { version = "2.0", features = ["keccak"] }
+hex = "0.4.3"
```
MartinquaXD: I'm not sure what the best practices for defining dependencies on a published crate are.
So far my understanding is that not setting a patch level version means the crate should be compatible with any of the patch level version whereas this would now only be compatible with versions > 0.4.3 which would possibly be a breaking change.
So conceptually I believe the best practice is only specifying major and minor version here since patch level releases are not expected to introduce breaking changes.
Also since rust resolves all sub-dependencies for all your crates and compiles everything with a shared set (if possible) upgrading those versions would be less important than in C++ for example.

ethcontract-mock/src/details/transaction.rs
```diff
@@ -2,6 +2,7 @@
 
 use ethcontract::{Address, H256, U256};
 
+#[allow(dead_code)]
```
MartinquaXD: Got rid of the warning by removing the fields.

ethcontract/src/lib.rs
```diff
@@ -165,6 +165,8 @@ pub mod dyns {
     pub type DynAllEventsBuilder<E> = AllEventsBuilder<DynTransport, E>;
 }
 
+#[allow(unknown_lints)]
+#[allow(clippy::mixed_attributes_style)]
```
MartinquaXD: Can be removed since I turned the hidden doc comments into regular comments since they are hidden anyway.

ethcontract-mock/src/details/transaction.rs
mateo-mro: I wasn't so sure I could delete it :sweat_smile:

ethcontract-common/Cargo.toml
mateo-mro: @MartinquaXD that's a really good question. I did some research and there is nothing clear, it really depends on what you want to achieve, stability, highest compatibility, etc.

Since Rust uses by default the caret sign for dependencies, it is totally fine to specify here `0.4`, because I can then use `0.4.3` when importing this without an issue.

.github/workflows/pull-request.yaml
mateo-mro: This update is required by the AWS stable release, otherwise it won't compile.

MartinquaXD: Overall I'm not a fan of bumping the rust version here. The version you are adding only requires 1.75 according to their readme (MSRV = stable - 2).
Also with the discussion on compatibility of sub-dependency version I think we should not specify the patch level version in our dependencies to give the dependency resolver more room to unify dependency versions.
Also since the MSRV definitely has to be increased (I agree that depending on a stable version of the aws crates makes sense) we should also update the readme of this repo since it mention MSRV as well.
And finally this needs to get published as a new version of the crate. Since we are increasing the MSRV I think to make it a new minor release instead of just doing a patch level release (i.e. 0.25.6 -> 0.26.0). This can be done by using `/ci/bump.sh`.

PR #971, opened 2024-05-14
Add more tuple implementations

Pzixel:
Add more tuple implementations (for tuples of length up to 21) since structures are allowed to have arbitrary amount of fields, they are represented as tuples by the crate and it's not uncommon to see some very large structures in the wild (intents can be quite big)


fleupold (approved):
PR description would be appreciated.

PR #975, opened 2025-01-06
Do not panic in the `kms::sign()` function

squadgazzz:
Removes unnecessary `unwrap()` call inside the `kms::sign()` function, which leads to unexpected behavior from the client side.

Also, includes workarounds in order to build the project with outdated dependencies.

### DIFF
--- a/ethcontract/src/transaction/kms.rs
+++ b/ethcontract/src/transaction/kms.rs
@@ -5,7 +5,6 @@
 //!
 //! It's quite hacky, however the hackiness does not leak outside this module.
 
-use aws_sdk_kms::operation::sign::SignOutput;
 use aws_sdk_kms::{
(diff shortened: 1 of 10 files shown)


ethcontract/src/transaction/kms.rs
```diff
@@ -86,8 +86,7 @@ impl Account {
             .signing_algorithm(SigningAlgorithmSpec::EcdsaSha256)
             .send()
             .await
-            .map_err(aws_sdk_kms::Error::from)
-            .unwrap();
+            .map_err(aws_sdk_kms::Error::from)?;
```
squadgazzz, line 89: The actual change.

MartinquaXD:
For some reason I can only comment but not approve.
Lgtm

PR #978, opened 2025-08-27
KMS account getters

squadgazzz:
This PR adds getters for KMS account, which is currently required for cowprotocol/services to migrate to alloy, where AwsSigner requires these fields: https://github.com/alloy-rs/alloy/blob/77d725b26321e1e3674b171c0788a13926870dd7/crates/signer-aws/src/signer.rs#L140-L147

Also contains changes in order to pass compilation on the latest rust versions.

### DIFF
--- a/ethcontract/src/transaction/kms.rs
+++ b/ethcontract/src/transaction/kms.rs
@@ -75,6 +75,16 @@ impl Account {
         self.address
     }
 
+    /// Returns the underlying KMS client.
+    pub fn client(&self) -> &Client {
+        &self.client
+    }
+
+    /// Returns the KMS Key ID.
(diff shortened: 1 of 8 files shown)


ethcontract/src/transaction/kms.rs
```diff
@@ -75,6 +75,16 @@ impl Account {
...
 
+    /// Returns the underlying KMS client.
+    pub fn client(&self) -> &Client {
+        &self.client
+    }
+
+    /// Returns the KMS Key ID.
+    pub fn key_id(&self) -> &str {
+        &self.key_id
+    }
```
squadgazzz, line 86: This is the main change.

MartinquaXD (approved):

PR #984, opened 2025-09-22
Exclude KMS account client from the Debug impl

squadgazzz:
KMS account client's debug information is too verbose, which is not useful. The proposal is to exclude this from the Debug impl entirely.

### DIFF
--- a/ethcontract/src/transaction/kms.rs
+++ b/ethcontract/src/transaction/kms.rs
@@ -23,7 +23,7 @@ use web3::{
 use crate::errors::ExecutionError;
 
 /// An AWS KMS account abstraction.
-#[derive( Debug)]
+#[derive(Clone)]
 pub struct Account {
     client: Client,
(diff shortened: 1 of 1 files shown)


jmg-duarte (approved):
LGTM, sad that Rust doesn't have something like [`#[debug(skip)]`](https://github.com/rust-lang/rust/issues/37009#issuecomment-2545591512) because for longer structs this would be kind of annoying to implement

MartinquaXD (approved):

MartinquaXD:
> LGTM, sad that Rust doesn't have something like https://github.com/rust-lang/rust/issues/37009#issuecomment-2545591512 because for longer structs this would be kind of annoying to implement

@jmg-duarte, it's not built in but this can be done with the derivative crate (see [here](https://mcarton.github.io/rust-derivative/latest/Debug.html#ignoring-a-field)).

